PR #6, opened 2021-09-02
Feature/doc recipe

meganrm:
**Pull request recommendations:**
This is starting to document the options for a recipe file

### DIFF
--- a/docs/RECIPE_SCHEMA.md
+++ b/docs/RECIPE_SCHEMA.md
@@ -92,10 +92,12 @@ _Required number_
 _Optional string_
 
 Location of publicly hosted packing result 
+
 ### use_periodicity
 _boolean_
 
 Whether to consider periodicity when packing objects. If `true` a packed object at the edge of the bounding will wrap to the other side of the bounding box. 
+
 ### EnviroOnly
 _boolean_
 
--- a/docs/RECIPE_SCHEMA.md
+++ b/docs/RECIPE_SCHEMA.md
@@ -119,23 +119,38 @@ An array that defines the names of directional gradients to use.
 _enum_. One of `"bhtree"`, `"jordan"`, `"sdf"`, `"jordan3"`, `"pyray"`, `"floodfill"`, `"binvox"`, `"trimesh"`, `"scanline"`. Defaults to `"jordan3"`.
 
 
- `"bhtree"`
+ `"bhtree"` build sthe compartment grid ie surface and inside point using bhtree.
  
- `"jordan"`
+ `"jordan"` builds the compartment grid ie surface and inside point using jordan theorem and host raycast.  Only computes the inner point. No grid.
+        This is independent from the packing. Help build ingredient sphere tree and representation.
+        - Uses BHTree to compute surface points
+        - Uses Jordan raycasting to determine inside/outside (defaults to 1 iteration, can use 3 iterations)
+        
 
-`"sdf"`
+`"sdf"` builds the compartment grid ie surface and inside point using signed distance fields
+        from the UT package.
+        
 
- `"jordan3"`
+ `"jordan3"` builds the compartment grid ie surface and inside point using jordan theorem and host raycast, with ray set to `3`.
  
  `"pyray"`
  
- `"floodfill"`
+ `"floodfill"` builds the compartment grid ie surface and inside point using flood filling algo from kevin. Takes a polyhedron, and builds a grid. In this grid:
+            - Projects the polyhedron to the grid.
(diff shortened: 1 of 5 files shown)


docs/RECIPE_SCHEMA.md
```diff
@@ -0,0 +1,319 @@
...
+
+### runTimeDisplay
+_boolean_
+
+### overwritePlaceMethod
+_boolean_
+
+### innerGridMethod
+_enum_. One of `"bhtree"`, `"jordan"`, `"sdf"`, `"jordan3"`, `"pyray"`, `"floodfill"`, `"binvox"`, `"trimesh"`, `"scanline"`
+ `"bhtree"`
```
meganrm, line 67: Same here: @gj210 if you could give 1 sentence descriptions of each (and which ones should be abandoned) that would be helpful. You can just do it as a rely in this comment.

docs/RECIPE_SCHEMA.md
```diff
@@ -0,0 +1,319 @@
...
+
+An array of two points that define the corners of the bounding box.
+
+### gradients 
+_Optional string[]_
+
+An array that defines the names of directional gradients to use.
+
+### smallestProteinSize
+_Required number_
```
meganrm, line 96: @gj210 are these required? (with largest protein size)

docs/RECIPE_SCHEMA.md
```diff
@@ -0,0 +1,319 @@
...
+
+_Optional boolean_
+
+### placeMethod
+
+_Optional enum_. One of `"RAPID"`, `"jitter"`, `"spheresBHT"`, `"pandaBullet"`
+
+Will be use if placeMethod isn't in an ingredient setup
+
+`"RAPID"`
```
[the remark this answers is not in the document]
gj210: _RAPID_  uses "Robust and Accurate Polygon Interference Detection (RAPID)"  http://gamma.cs.unc.edu/OBB/ to test if the ingredient being placed in the packing loop is colliding with any neighboring objects fed into the list of neighbors.

_jitter_ uses a simple algorithm developed by GJ, M-A-A, MS, and LA to test if a single sphere, sphere-tree, or other primative (box and cylinder) are colliding with masked/unallowed points on the grid... I can't recall all of the allowed types, but check the input parameters that it accepts.  I believe there is also an option to perform simple collisions directly between primatives, e.g. sphere-sphere, sphere-box, sphere-cylinder, cylinder-box, and spheretree-others

_spheresBHT_ gets the sphereTrees of the potential colliding neighbors, and does an efficient sphereTree-sphereTree collision detection of the sphereTree for the object being packed against the sphereTrees of each neighbor- returns false if a collision is detected

_pandaBullet_ Python wrapper for Bullet Physics Engine (popular in ~2010 and used by C4D, Maya, Blender, etc) that provides a variety of object-object collision detection, including collision min/max overlap distance, etc.  Allows relaxation in its own loop, springs, rejection, meshes, primitives, etc.

docs/RECIPE_SCHEMA.md
gj210: This is not required, but is a hack to allow the user to manipulate the voxel size of the MasterGrid instead of accepting the calculated defaults... e.g. a sparse grid will lead to very fast but potentially more sparse packing, and allows unexpected things like for an object to jitter beyond its pickedPoint, which is effectively an inefficient way to pack. It can be helpful to force more attempts when packing densely, e.g. odd-shaped lipids into a lipid bilayer, or for sphereTrees that are overly-coarse

gj210: If deleted, then this value will be calculated correctly by the algorithm, based on the spheretree input file (packing radius) or the packingRadius calculated from ingredients that use primatives for their geometries.

docs/RECIPE_SCHEMA.md
gj210: This will require some work to get these right.  Jordan3 may be our best "fast" option for now.  I can't recall, but "sdf" may be the brute-force 100% guaranteed not to leak/incredibly slow option. "trimesh" may be the option either Kevin or Megan worked on with other Berkeley students.

gj210 (commented):
Did best to annotate the requested lines off the top of my head. Some should be confirmed.

docs/RECIPE_SCHEMA.md
meganrm: I looked through some of these comments and added them in. But some of them look like the function was just copied and nobody updated the comment, so I'm not sure if they still apply

PR #7, opened 2021-09-02
Feature/convert to simularium

meganrm:
## Problem 
I needed the newer release of simulariumio to add in rotations.

## Solution

This updates the 1convert to simularium1 code to use the newer release of simulariumio and include rotations. 

I also updated it to be able to convert "cellpack version 1" results, which are results that have "compartments" instead of "cystoplasme" as a key.


cellpack/bin/simularium_converter.py
```diff
@@ -3,12 +3,19 @@
 import traceback
 
 # import cellpack.autopack.transformation as tr
-from collections import OrderedDict
 import numpy as np
 import json
 import logging
+from scipy.spatial.transform import Rotation as R
```
meganrm, line 9: this is the only file that has actual code changes, everything else is formatting fixes

PR #9, opened 2021-09-21
Jitter packing working with sphere trees and periodicity 

meganrm:
Note: I branched this off the cleanup branch, so I'm doing the PR to that until it gets merged so we can see the true diff. 

# Problem: 
Jitter packing was allowing overlaps, didn't do closePartner packing, and didn't do periodic collisions. 
Also, the sphere tree collision check was saving insidePoints based on the encapsulating radius so wouldn't allow close packing to happen. 

# Solution:
1. Updated `jitter_place` to do correct distance check for placing ingredient
2. Updated sphere tree walk to return recursive check
3. Updated sphere tree walk to jump all the way down the sphere tree to calc inside points if no collisions have been found and the ingredient is being place. 
2. Make sure the collisions detected from periodic check are added to the new insidePoints and newDist dictionaries, including corner points. Removed unneeded extra loop to update distances, speeds up the whole loop by almost 2x.
3. Cleaned up Grid and BaseGrid functions (one `getPointsInCube` just called `getPointsInSphere`
4. Moved the getPointToDrop and associated functions to the Ingredient class, since it's really a per ingredient call. I also factored them out for clarity and added comments. 
5. Move the plotly results to the analysis class. 

# Results
## Before changes
 jitter place with random place (allowing overlaps) and closePartner packing (spheres should only be touching other spheres):
Screen Shot 2021-09-20 at 6 32 42 PM
SphereTree file (not allowing overlaps with encapsulating radius for ingredients that were placed early in the packing): 
Screen Shot 2021-09-16 at 8 22 30 PM

#### jitter random place (no overlaps):
Screen Shot 2021-09-20 at 6 34 47 PM

#### jitter closePartner (packing correctly)
Screen Shot 2021-09-20 at 6 35 32 PM

#### using a sphereTree and with periodic collisions.
Screen Shot 2021-09-21 at 12 47 44 PM


cellpack/autopack/Ingredient.py
```diff
@@ -5022,9 +5144,10 @@ def jitter_place(
             )
             periodic_collision = False
             collision_results = []
+
             if periodic_pos is not None and self.packingMode != "gradient":
                 for p in periodic_pos:
-                    periodic_collision = self.collision_jitter(
+                    periodic_collision, insidePoints, newDistPoints = self.collision_jitter(
```
meganrm, line 5150: these points were actually being saved by update distances

cellpack/autopack/Ingredient.py
```diff
@@ -5106,26 +5242,10 @@ def jitter_place(
                 ),
             )
             env.timeUpDistLoopTotal += time() - timeUpDistLoopStart
-            if periodic_pos is not None and self.packingMode != "gradient":
```
meganrm, line 5109: this is redundant with the collision check

cellpack/autopack/Ingredient.py
```diff
@@ -3241,86 +3340,81 @@ def checkSphCollisions(
                     self.vi.updateBox(box, cornerPoints=bb)
                 self.vi.update()
 
-            pointsInCube = histoVol.grid.getPointsInCube(
-                bb, posc, radius_of_area_to_check, info=True
+            pointsToCheck = histoVol.grid.getPointsInSphere(
```
meganrm, line 3343: getPointsInCube was calling getPointsInSphere, so I just directly call it here

cellpack/autopack/Ingredient.py
```diff
@@ -3241,86 +3340,81 @@ def checkSphCollisions(
...
-                ):
-                    # if there is an object encroaching into this packing area, that is
-                    # closer to the packing location than the radius of the object
-                    # we are trying to pack, probably a collision
-                    if level < self.maxLevel and (level + 1) < len(self.positions):
-                        nxtLevelSpheres = self.positions[level + 1]
-                        nxtLevelRadii = self.radii[level + 1]
+                collision = (
+                    distance[pt] + distance_to_packing_location
+                    <= radius_of_ing_being_packed
```
meganrm, line 3361: This is the change that actually prevents collisions. We need to use the distance to the closes surface to determine if the closes surface to too close to the packing location we're trying

cellpack/autopack/Ingredient.py
```diff
@@ -3241,86 +3340,81 @@ def checkSphCollisions(
...
-                            cradii.append(nxtLevelRadii[sphInd])
-                        collision = self.checkSphCollisions(
-                            ccenters,
-                            cradii,
+                        # ccenters = []
+                        # cradii = []
+                        # for sphInd in self.children[level][sphNum]:
+                        #     ccenters.append(nxtLevelSpheres[sphInd])
+                        #     cradii.append(nxtLevelRadii[sphInd])
+                        return self.checkSphCollisions(
```
meganrm, line 3379: it's import to return this recursive call so we can keep walking down the tree, instead of just calling it once.

cellpack/autopack/Ingredient.py
```diff
@@ -3329,14 +3423,28 @@ def checkSphCollisions(
...
-                else:
-                    continue
-
+            if not at_max_level:
+                # we didn't find any colisions with the this level, but we still want
+                # the inside points to be based on the most detailed geom
+                new_level = self.maxLevel
+                nxtLevelSpheres = self.positions[new_level]
+                nxtLevelRadii = self.radii[new_level]
+                return self.checkSphCollisions(
```
meganrm, line 3437: This jumps us all the way down to the more detailed geom to calculate the inside points. Before if no collisions had been found, the inside points would be based on the encapsulating radius.

cellpack/autopack/Ingredient.py
```diff
@@ -5001,11 +5110,24 @@ def jitter_place(
...
         insidePoints = {}
         newDistPoints = {}
+
         for attempt_number in range(self.nbJitter):
             packing_location, _, _, _ = self.randomize_translation(
                 env, targeted_master_grid_point, rot_mat
             )
             jitter_rot = self.randomize_rotation(rot_mat, env)
+
+            if env.ingrLookForNeighbours and self.packingMode == "closePartner":
```
meganrm, line 5120: Add this in for partner packing. But ideally, this is a more general call that doesnt have to be in each different place call.

cellpack/autopack/Ingredient.py
```diff
@@ -3241,86 +3341,81 @@ def checkSphCollisions(
...
-                        return True, insidePoints, newDistPoints
+                        self.log.warning("checked pt that is not in container")
+                        return True, {}, {}
                     if pt in insidePoints:
-                        # NOTE: Don't we want the number closer to zero? Not the most negative?
-                        if signed_distance_to_sphere_surface > 0:
-                            signed_distance_to_sphere_surface = - signed_distance_to_sphere_surface
-                        if signed_distance_to_sphere_surface < insidePoints[pt]:
+                        if abs(signed_distance_to_sphere_surface) < abs(
+                            insidePoints[pt]
```
meganrm, line 3417: This check seemed incorrect to me before, since they are both negative numbers and we want the one that tells us the closes surface, not the most negative number. I uses `abs` just so that it's consistent with the `min` check for distance below.

cellpack/autopack/Ingredient.py
```diff
@@ -2067,7 +2066,7 @@ def getSpheres(self, sphereFile):
             centers.append(cl)
             radii.append(rl)
             children.append(ch)
-
+        print(rmin, rmax, centers, radii, children)
```
meganrm, line 2069: get rid of this print statement

cellpack/autopack/Ingredient.py
meganrm: do we want to have the other check as a "soft" packing option?

cellpack/autopack/Ingredient.py
```diff
@@ -5118,26 +5254,10 @@ def jitter_place(
...
-                            nbFreePoints,
-                            distance,
-                        ),
-                    )
-
             # add one to molecule counter for this ingredient
             self.counter += 1
             self.completion = float(self.counter) / float(self.nbMol)
-
+            self.update_data_tree(packing_location, rot_mat, ptInd)
```
meganrm, line 5260: should only do this if we're doing partner packing

PR #12, opened 2021-10-11
Feature/use classes

meganrm:
### Code organization problems
1. A lot of functions in Environment and Ingredient had if/else statements checking for what type of ingredient it is.
2. There was a lot of code in the Environment class that almost entirely dealt with instances of the Ingredient class
3. The update distances functions were on the Environment class but they never referenced the env
4. Env was unpacking or setting ingredient level variables, and then sending those into an ingredient method. 
### Code functionality Problems
1. Panda bullet wasn't use sphere trees to do collision
2. The grid was being filtered and masked for each ingredient each time by the same cuttoff distance 

### Organization solutions
1. I moved code that did the same thing, but with different implementations based on the type the ingredient was into the respective class for the ingredient as a method with the same name. Ie, now each ingredient has a "jitter_collision" instead of "jitter_collision" having a switch on "ingr.Type" and then calling a different function. Another example: instead of setting up this dictionary, [each ingredient class now has it's own `create_rb_node`](https://github.com/mesoscope/cellpack/commit/e6a1359a682662a2b44429d73974840575ca867d)
2. I moved functionality down to the Ingredient class if the function was dealing with an ingredient more than the env.
3. I made the update distances functions into static methods on the grid class
4. Only send things into the ingredient method that the ingredient doesn't already know. Removed a number of arguments to function calls. [Example, the ingredient already knows it's current compartment id and radius, env shouldn't send these values in](https://github.com/mesoscope/cellpack/commit/cdb9997f1f8f7876c2a86d55917e86e1cd60f02d)

### Functionality solutions
1. Made sure the create rigid body representation code was using the most detailed geometry 
2. Use stored points on each ingredient to start from whenever checking for a new point to drop at. This saves some compute time. 

Screen shots, pandaBullet and jitter now pack exactly the same:
newplot (17)
newplot (18)

newplot (15)
newplot (16)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -43,7 +43,7 @@ requirements = [
     "panda3d==1.10.9",
     "pmw==2.0.1",
     "scipy==1.6.2",
-    "simulariumio==1.1.0",
+    "simulariumio>=1.2.0",
     "pycollada==0.7.1",
 ]
 
@@ -73,7 +73,7 @@ setup(
     entry_points={
         "console_scripts": [
             "analyze=cellpack.bin.analyze:main", 
-            "sim-convert=cellpack.bin.simularium_converter:main"],
+            "convert=cellpack.bin.simularium_converter:main"],
     },
     install_requires=requirements,
     license="MIT license",


meganrm:
> Really impressive work Megan. So glad to see the sphere tree at use ! I am surprised that both bullet and jitter will get the same timing. Will be interesting to see how theses number evolve as you generate bigger scene.

Thanks. Yeah, for the sphere trees with this size box, bullet maxed out at 37 packings, while jitter can place 5 additional ingredients and therefore does take longer.  The graphs I posted above had the number to pack set at 37 for the direct comparison. 
newplot (19)

PR #22, opened 2021-10-29
Fix/test compartment

corredD:
Testing  a simple recipe build from mesoscope and used in the tutorial http://www.autopack.org/documentation/tutorials/nbcr-summer-training-2018
I started to fix all issues when running  analyze -r test-recipes\myrecipe.json -o output2 -d 3
I stop when encountering issues with the sphereTree level.


corredD:
to test run :  
`analyze -r test-recipes\myrecipe.json -o output2 -d 3`

cellpack/autopack/BaseGrid.py
```diff
@@ -193,6 +193,8 @@ def __init__(
             # what about collision ?
 
     def setup(self, boundingBox, space):
+        if space == 0:
+            space = 20
```
meganrm, line 197: Can you explain the reason for this? Why 20?

meganrm:
can you rename `myrecipe` to `test_compartment` so it's clear what this test recipe is for?
Also, can you explain these settings:
`comp1: {
    "geom_type": "mb",
    "geom": "mb",
    "thickness": 7.5
...
}`
Is this a primitive or a mesh?
Is `mb` a keyword?  What are the other options? 
What's the difference between `geom_type` and `geom`? Do both need to be set?

cellpack/autopack/BaseGrid.py
corredD: it was quicker to do that rather that track down the automatic value calculated based on the ingredient size. So the grid would not be huge for the test. So it needs more testing.

corredD:
> can you rename `myrecipe` to `test_compartment` so it's clear what this test recipe is for? Also, can you explain these settings: `comp1: { "geom_type": "mb", "geom": "mb", "thickness": 7.5 ... }` Is this a primitive or a mesh? Is `mb` a keyword? What are the other options? What's the difference between `geom_type` and `geom`? Do both need to be set?

I will rename it.
The geom_type defined the type of geometry used for the compartments, and geom hold the actual data which can be of different type of course. "mb" refer to metaballs where primitives are used to build a isosurface from the signed distance function of the metaballs or primitives. geom_type can refer to "file" or "raw" wich will be either .obj file or the actual vertices, triangle and normal of the mesh. This also can be worked on, how a compartments should be defined and saved.

cellpack/autopack/BaseGrid.py
meganrm: can you add a "TODO:" comment here? this doesn't seem like a good thing to leave in production code.

meganrm (approved):

Issue #23, opened 2021-10-29
Conversion code incorrectly assumes recipes have either a cytoplasm OR compartments 


meganrm:
## Description
If you convert a recipe/result file that has both ingredients in the cytoplasm and compartments it only shows the ingredients in the cytoplasm 

## Expected Behavior
it should show all the ingredients 

## Reproduction
use test recipe that Ludo added for compartments

## Environment
*Any additional information about your environment*

PR #25, opened 2021-11-11
Fix/simularium convert

meganrm:
Problem
=======
Simularium convert wasn't unpacking both cytoplasm and compartment #23 
The scaling was off for the simularium viewer
Also, simulariumio now allows for geometry to be specified, which we weren't using

Solution
========
Created a slimmed down `io_tools` that isn't coupled with the ENV and just reads in the recipe. 
Added a scale factor
Added a geometry cli argument to get the right url for available geometry. 
**NOTE**: two known issues: bounding box is off, rotations for blender are off by 90 degrees
@corredD updated the ingredient code to support splines

## Type of change

Steps to Verify:
----------------
1. After activating env
2. `convert -r [PATH_TO_CELLPACK]/cellpack/cellpack/test-recipes/NM_Analysis_FigureC1.json -p [PATH_TO_CELLPACK]/cellpack/cellpack/test-recipes/NM_Analysis_FigureC1.4.cpr.json`
3. Recipe should look like 
Screen Shot 2022-01-06 at 11 30 04 AM

-----------------------
1. main file/entry point
2. other important file

Thanks for contributing!


meganrm:
Screen Shot 2021-11-11 at 10 28 44 AM
@corredD I need help with the rotation. This is a conversion of the HIV 1-6 result, and the matrix protein is clearly not oriented correctly. I'm doing the rotation conversion using the scipy package.


```python
def get_euler_from_matrix(self, data_in):
        rotation_matrix = [np.array(data_in[0][0:3]), np.array(data_in[1][0:3]), data_in[2][0:3]]
        return R.from_matrix(rotation_matrix).as_euler("xyz", degrees=True)
```

meganrm:
[HIV.simularium.zip](https://github.com/mesoscope/cellpack/files/7522816/HIV.simularium.zip)
To look at it yourself, drag this file onto the viewer at https://allen-cell-animated.github.io/simularium-viewer/
(unfortunately the "hide all" button isn't working, so you have to click each one off to just look at the matrix protein alone.

corredD:
There is still an issue with the OBJ loading in simularium in term of rotation.
It should look like this : 
image
I found the issue with the OBJ files, the default export option of blender was switching the Y and Z. The proper way to export a geometry as a wavefront OBJ from Blender is to use Y-forward and Z-up. Looks better now in simularium : 
image

corredD:
I did a commit that replace the scipy rotation and change the scaling factor to 1 because simularium doesnt apply the scaling factor when loading the OBJ.

cellpack/bin/simularium_converter.py
```diff
@@ -37,7 +39,7 @@ class ConvertToSimularium(argparse.Namespace):
     DEFAULT_OUTPUT_DIRECTORY = "/Users/meganriel-mehan/Dropbox/cellPack/"
     DEFAULT_INPUT_RECIPE = "/Users/meganriel-mehan/dev/allen-inst/cellPack/cellpack/cellpack/test-recipes/NM_Analysis_FigureC1.json"
     DEFAULT_GEO_TYPE = "OBJ"  # Other options: SPHERE or PDB
-    DEFAULT_SCALE_FACTOR = 1.0 / 10
+    DEFAULT_SCALE_FACTOR = 1.0 # / 10
```
meganrm, line 42: we need this scaling for pdbs. Was it causing any problems as is? If not it would be better to leave it. If so then we should adjust it if the GEO_TYPE is PDB, otherwise this is now broken for other recipes

cellpack/bin/simularium_converter.py
```diff
@@ -237,7 +243,7 @@ def unpack_curve(self, data, time_step_index, ingredient_name, index, agent_id):
             self.max_fiber_length = len(data[curve])
 
     def unpack_positions(
-        self, data, time_step_index, ingredient_name, index, agent_id, comp_id=0
+        self, data, time_step_index, ingredient_name, index, agent_id, comp_id=0, left_hand=False
```
meganrm, line 246: is this only for blender OBJs? does it work with meshes exported from other 3D software?

corredD: sorry I forgot to remove it. The problem was indeed from the blender export and should not be the case in other software. 
The scaling is an issues, and you should be able to pass the scaling factor to simularium. so that simularium does the scaling not the converter. Assuming you want both geom and PDBs, for instance a mesh for the compartment and PDBs for the ingredient, you want to be able to have the PDb and the geom at the same scale.

corredD: otherwise could the scaling factor be an argument to the convert command line ?

meganrm: I'm confused by the scaling. The OBJs I had before were working with the .1 scale factor. Now with a scale factor of 1 the pdbs are way to small, and the OBJs work, but the whole scene is also way too large for the simularium viewer (the objects are disappearing in the back) did you scale up these OBJ files?

meganrm: shouldn't this be internally consistent in cellPACK?

cellpack/bin/simularium_converter.py
```diff
@@ -202,11 +204,15 @@ def get_ingredient_data(self, cytoplasm, main_container, ingredient):
...
+        # rotation_matrix = [data_in[0][0:3], data_in[1][0:3], data_in[2][0:3]]
+        # euler = R.from_matrix(rotation_matrix).as_euler("XYZ", degrees=False)
+        euler2 = TR.euler_from_matrix(data_in)
+        return euler2
 
     def get_euler_from_quat(self, data_in):
-        return R.from_quat(data_in).as_euler("XYZ", degrees=False)
+        # euler = R.from_quat(data_in).as_euler("XYZ", degrees=False)
+        euler2 = TR.euler_from_quaternion(data_in)
+        return euler2
```
meganrm, line 215: with these new changes the HIV rotations no longer look right. I would prefer to have the HIV/PBD working then trying to get blender OBJs to work because from the cellPACK chart blender is the most annoying export anyways

meganrm: (also, just a note on commits..... remember to just write the code as you want it (`euler` instead of `euler2` and don't comment out code, just delete it). It's easy enough to revert commits if we don't want the change and if we do want it we dont have to do the extra work of cleaning up the code later, or being confused why it looks this way.

cellpack/bin/simularium_converter.py
corredD: I scaled them temporarly by 0.1 to make sure simularium was not scaling them, and then revert back to default scaling. So the problem here is simularium not being aware of the scaling factor used during the conversion. The rotation issues is a deeper problem on how it was saved. If it was saved left-hand/righthand and if it was saved using transformation.py. If you want to use scipy, change the saving code that generate the quaternion so its consistent. transformation.py and scipy gave different results during the euler<->quaternion convertion for the same axis,angle rotation. because the quaternion order is different x,y,z,w in scipy and likely w,x,y,z in transformation.py.

PR #26, opened 2021-11-12
Fix/convert bacteria recipie

meganrm:
Problem
=======
Rotations were being inconsistently converted

Solution
========
Using a simpler recipe, to test rotations, and had to fix some issues with the "grow" ingredient to make sure the recipe was working. 

The rotations are still weird for this particular recipe, but have verified it that it's a blender issue. 

with @corredD 

## Type of change


cellpack/autopack/ingredient/Ingredient.py
```diff
@@ -2975,16 +2978,14 @@ def attempt_to_pack_at_grid_location(
             )
 
         # grow doesnt use panda.......but could use all the geom produce by the grow as rb
-        if self.placeType == "jitter" or self.Type == "Grow" or self.Type == "Actine":
+        if self.Type == "Grow" or self.Type == "Actine":
```
meganrm, line 2981: did you mean to remove the placeType check here? If placeType is jitter, we should call jitter_place

corredD: the growing ingredient (Grow.py)  is a special ingredient that use its own placing function that deals with the placing type. To avoid confusion I renamed it to grow_place, while before it was overwriting the jitter_place function of the parent class Ingredient.

cellpack/autopack/ingredient/grow.py
```diff
@@ -2574,9 +2581,9 @@ def add_rb_multi_sphere(self):
             )  #
         return inodenp
 
-    def jitter_place(
+    def grow_place(
```
meganrm, line 2584: I don't think this should be renamed. It makes sense for different classes to have their own implementation of the same function. So if placetype == 'jitter' whatever implementation of jitter that ingredient has, it will use. That's the whole point of putting it in it's own class

cellpack/autopack/ingredient/Ingredient.py
meganrm: I haven't pulled this down, but from the diff on here, it now looks like there is no way to call `jitter_place`. I think it makes more sense to have this just be (pseudocode) `if self.place == "jitter" call self.jitter_place` and then each ingredient has its own implementation of jitter (defaults to spheres). That way we can get rid of the `if self.Type == "Grow" or self.Type == "Actine"` check here completely

cellpack/autopack/ingredient/multi_cylinder.py
```diff
@@ -359,3 +386,145 @@ def add_rb_node(self, worldNP):
             )  # math.sqrt(s), 1)# { XUp = 0, YUp = 1, ZUp = 2 } or LVector3f const half_extents
             inodenp.node().addShape(shape, TransformState.makeMat(pMat))  #
         return inodenp
+
+    def checkCylCollisions(
```
meganrm, line 390: this looks like a duplication of `collision_jitter`. The idea of splitting these into classes with their own methods is to reduce the amount of duplicate code. It should just be calling `collision_jitter`

meganrm:
A few additional comments questions on this change set overall as it stands (it seems like you're still working on it too). 
1. There are a lot of new binary files under `mgl_tools`. What are they? A goal is to remove mgl_tools altogether since it's not really being maintained, and you can't pip/conda install it for python 3. Do we need these new files?
2. Can you remove all results files and large recipes from the repo? It makes sense to have a couple test recipes that we can use for specific functionality, but this shouldn't just be reproducing the git database. When you delete these files, it would be good to also delete them from the git history, otherwise it's always going to be storing a copy of them and it will make the history larger than it needs to be `git rm --cached path/to/file`

cellpack/autopack/ingredient/multi_cylinder.py
```diff
@@ -335,6 +337,31 @@ def collision_jitter(
             cylNum += 1
         return False, insidePoints, newDistPoints
 
+    def pandaMatrice(self, mat):
```
meganrm, line 340: this already exists on env (which may not be the best place for it) but we shouldn't just have the same exact function here. Probably makes more sense to make `pandaMatrix` into a static method on env since it's not using `self` at all

cellpack/autopack/plotly_result.py
```diff
@@ -16,6 +16,8 @@ def update_title(self, title):
 
     @staticmethod
     def format_color(color):
+        if color is None:
+            color = [1.0, 0.0, 0.0]  # RED default
```
meganrm, line 20: couldn't you just do `def format_color(color=[1.0, 0.0, 0.0]): # RED default` ?

cellpack/bin/simularium_converter.py
```diff
@@ -201,17 +238,19 @@ def get_ingredient_data(self, cytoplasm, main_container, ingredient):
...
-        ).transpose()
-        return R.from_matrix(rotation_matrix).as_euler("xyz", degrees=True)
+        rotation_matrix = [
+            np.array(data_in[0][0:3]),
+            np.array(data_in[1][0:3]),
+            data_in[2][0:3],
+        ]
+        return np.degrees(
+            tr.euler_from_matrix(rotation_matrix)
+        )  # R.from_matrix(rotation_matrix).as_euler("xyz", degrees=True)
```
meganrm, line 248: this may be a good idea, but I tried this and it still seems like the rotations are incorrect

cellpack/bin/simularium_converter.py
```diff
@@ -282,67 +324,71 @@ def get_euler(self, data_in):
...
+                for j in range(len(ingredient_results_data["results"])):
+                    self.unpack_positions(
+                        ingredient_results_data,
+                        time_step_index,
+                        ingredient_key,
+                        j,
+                        self.agent_id_counter,
+                    )
+                    self.agent_id_counter = self.agent_id_counter + 1
+            elif ingredient_results_data["nbCurve"] > 1000000:
```
meganrm, line 354: this doesn't seem right, why greater than `1000000`?

corredD:
> A few additional comments questions on this change set overall as it stands (it seems like you're still working on it too).
> 
> 1. There are a lot of new binary files under `mgl_tools`. What are they? A goal is to remove mgl_tools altogether since it's not really being maintained, and you can't pip/conda install it for python 3. Do we need these new files?
Since the .so are in the repo, I put the windows pyd as well. But we should get rid of all of them for sure if the goal is to remove mgl_tools. 
> 2. Can you remove all results files and large recipes from the repo? It makes sense to have a couple test recipes that we can use for specific functionality, but this shouldn't just be reproducing the git database. When you delete these files, it would be good to also delete them from the git history, otherwise it's always going to be storing a copy of them and it will make the history larger than it needs to be `git rm --cached path/to/file`
yes

cellpack/autopack/ingredient/multi_cylinder.py
corredD: there is 3 `collision_jitter ` functions. One for sphere in ingredient.py. One for cylinder in multi_cylinder, and one for Cube in single cube. So if we call collision jitter from the parent class, this will call the sphere test and not the cylinder test.

cellpack/autopack/ingredient/Ingredient.py
corredD: To make it clear I separate the grow ingredient and the rest, since the grow ingredient deal itself with the different type methods.

cellpack/bin/simularium_converter.py
corredD: the euler angle calculated here are the correct one, and show up correctly in Blender. Do you know the coordinate system used in simularium ? Y is up ? Z is forward or backward ?

cellpack/autopack/ingredient/multi_cylinder.py
meganrm: the way I was setting this up if the ingredient is an instanced of the mutil_cylinder class, it should use the `collision_jitter` on its class. They're each supposed to have their own implementation of the same method. The one on `Ingredient.py` could be moved to sphere, so there is no default, either way the methods should get overwritten by the subclass, you know for sure it's not working correctly?

cellpack/bin/simularium_converter.py
corredD: Sorry I didnt wanted to debug the curve and wanted to skipped them. I put a TODO comment.

cellpack/autopack/ingredient/multi_cylinder.py
corredD: Yes there is duplication of code in checkCylCollisions and collision_jitter. I added checkCylCollisions  because it was called in Grow.py, realized after its the same code of collision_jitter. I think I made proper change in the latest push.

meganrm: ok cool, yeah, I just realized going through this that I never refactored the `jitter_place` and collision methods on the grow ingredient to match the refactor I did to the other methods. Overall I've been trying to reduce duplicate code because a lot of the different place methods all had the same setup lines and were doing a lot more than just finding if the point can be packed or not. I think the changes you have are ok, but also, since this is just to get the conversion to work, they're not really necessary  here either.

cellpack/bin/simularium_converter.py
meganrm: Screen Shot 2021-11-15 at 6 08 19 PM

meganrm: so the camera is in the positive Z direction in this view

meganrm: this is definitely an improvement since they all seem off in the same way

corredD: the problem is either in the 3d geometry export/import or in how the rotation is applied. Because the euler passed on this  model should be all (0,0,z) since its a 2d packing.

meganrm: Ok, I got it to look like the picture you posted in blender, but I had to hard code the x and z rotations. One thing I realized is that simularium is actually in radians. I think it was once in degrees, but either way that took longer than it should have for me to realize. I think simularium is also importing the geometry incorrectly: if I send in 0,0,0 for the rotation, the meshes are all standing up in Z, so I had to first rotate them 90 degrees to get them to lay flat. I also haven't figure out why the bounding box is off for this recipe, i haven't seen that in other recipes I've converted and looking at the positions, it seems like they aren't shifted or anything.

corredD: I dig up a little more that particular recipe, and realize it was setup in left hand coordinate system and the collada files were coming from Cinema4D which is left hand. The ingredient actually is defined in the lefthand coordinate system for the mesh. When I exported the file in obj and then reloaded them in blender, it automatically rotate 90degree on X. So it was both the .obj and the viewer ! This recipe should be deprecated, and I will try to prepare a cleaner one.

meganrm: awesome, yeah, thanks for digging in.... I just pushed a commit that has it using radians but not messing with any of the coordinates systems

cellpack/autopack/upy/dejavuTk/dejavuHelper.py
```diff
@@ -19,15 +18,13 @@
...
+# standardmodule
 import os
 import numpy
 from PIL import Image
-import collada
-
-from cellpack.autopack.upy import hostHelper
 
 # DejaVu module
+from cellpack.autopack.upy import hostHelper  # its confusing that upy is in three different place
```
meganrm, line 27: I'm trying to delete mgl tools, so I'm copying any part of upy that I need into thee autopack module. Is that what you mean about 3 different places?

cellpack/autopack/BaseGrid.py
```diff
@@ -161,7 +164,7 @@ def __init__(
         # object (or an compartment surface NOT IMPLEMENTED)
         self.distToClosestSurf = []
         self.distToClosestSurf_store = []
-
+        print(space)
```
meganrm, line 167: if you need a log, can you use the logger? Otherwise delete this

cellpack/autopack/Compartment.py
```diff
@@ -294,22 +296,6 @@ def getDejaVuMesh(self, filename, geomname):
 
         return geom
 
-    def saveDejaVuMesh(self, filename):
```
meganrm, line 297: was this just not being used or was it broken?

cellpack/autopack/Environment.py
```diff
@@ -1334,7 +1334,8 @@ def setMinMaxProteinSize(self):
             if maxi > self.largestProteinSize:
                 self.computeGridParams = True
                 self.largestProteinSize = maxi
-
+        print(self.smallestProteinSize)
+        print(self.largestProteinSize)
```
meganrm, line 1338: logger or delete

cellpack/autopack/IOutils.py
```diff
@@ -487,14 +487,14 @@ def addCompartments(env, compdic, i, io_ingr):
...
-                (rep != "None" and len(rep) != 0 and rep != "" and rep == ""),
-            )
+            # print(
+            #    "rep ?",
+            #    name,
+            #    geom,
+            #    rep,
+            #    rep_file,
+            #    (rep != "None" and len(rep) != 0 and rep != "" and rep == ""),
+            # )
```
meganrm, line 497: you can just delete these too if they're not useful

cellpack/autopack/upy/hostHelper.py
```diff
@@ -5028,6 +5030,26 @@ def writeToFile(self, polygon, filename):
             filename, verts=vertices, faces=faces, vnorms=vnormals, fnorms=fnormals
         )
 
+    @classmethod
+    def saveDejaVuMesh(self, filename, verts=[], faces=[]):
```
meganrm, line 5034: I think that the convention for classmethods is that the first argument is `cls` instead of `self1

PR #33, opened 2022-03-15
Fix/remove bhtree

meganrm:
Problem
=======
Need to remove bhtree so the project can be built on any system 

Solution
========
Using scipy package tree algorithm.

In replacing the algorithm we figured out that the search tree wasn't actually transversing the sphere tree levels. So we update the collision check with a while statement to go down the levels. This now packs sphere trees faster than jitter, but with the exact same result. 
with @mogres and @prismofeverything 
## Type of change

**Before**
newplot (20)
**After**
newplot (23)

**Jitter, for comparison**
newplot (22)


cellpack/autopack/ingredient/Ingredient.py
```diff
@@ -2491,48 +2443,65 @@ def np_check_collision(self, position, rotation):
...
                 continue
-            d, i = o.OGsrfPtsBht.query(position)
+            d, i = o.OGsrfPtsBht.query(packing_location)
             if d < self.encapsulatingRadius * 1.1:
-                p = self.get_new_pos(self, position, rotation)
-                d, i = o.OGsrfPtsBht.query(p)
-                D = d - numpy.array(self.radii[0])
-                nb = numpy.nonzero(D < 0.0)[0]
-                if len(nb) != 0:
+                pos_of_attempting_ingr = self.get_new_pos(
```
meganrm, line 2498: We need to do the same changes here to use positions and rotations at a certain level instead of `[0]`

mogres: Is this why we ended up getting some overlaps with the last iteration?

meganrm: no, I just figured out why the overlaps were happening: the radii of the packed ingr wasn't in the same order as the results from `query` so the distance check was sorta random. 

The last commit has these changes. 

newplot (24)
newplot (25)

PR #50, opened 2022-07-14
Feature/mesh

meganrm:
Problem
=======
This makes it so we can have a mesh compartment. #15 
NOTE: this looks like a big PR because I deleted the mgl_tools folder

Solution
========
The main big change is the creation of a Mesh store, which holds all the functionality around loading and manipulating a mesh. I am mostly wrapping [Trimesh.py](https://trimsh.org/trimesh.html) in that class. 

Other changes:

- Updated the inner grid functions to accurately sort the grid points into inside, surface, and outside points. Currently, `trimesh` and `raytrace` work. I also switched from a bounding box check to a query ball check to gather the possible inside points which speeds up the calculation. 
- I added a grid visualization to the simularium output that puts little colored spheres at the grid points. 

## Type of change

Steps to Verify:
----------------
1. run `analyze -p spheresSST -np -r /cellpack/test-recipes/test_compartment.json -d 3`
2. drop both the resulting file and `membrane_1.obj` on the viewer to see the packing and grid points

Screenshots (optional):
-----------------------
Screen Shot 2022-07-14 at 12 53 10 PM


cellpack/autopack/Analysis.py
```diff
@@ -475,7 +475,6 @@ def rectangle_circle_area(self, bbox, center, radius):
                 area = self.g.get_rectangle_cercle_area(
```
meganrm, line 475: main change here is generalizing the 'show plotly plot' command line option to show the grid spheres if the packing is in 3d

cellpack/autopack/BaseGrid.py
```diff
@@ -135,11 +135,10 @@ def updateDistances(
         return nbFreePoints
```
meganrm, line 135: Very minor changes in this file, just changing `space` to `spacing`

cellpack/autopack/ingredient/grow.py
```diff
@@ -1599,8 +1599,7 @@ def grow(
         k = 0
```
meganrm, line 1599: Just formatting changes

mogres:
The module `rtree` is required to run `analyze`. Could we add it to the setup? Lmk if this is added already and I just have to re-run the setup.

image

mogres:
Possible suggestion: could we copy the required `.obj` files to the output folder after packing? It would make it easier to drag and drop everything together onto the simularium viewer. Not sure how feasible this is for larger files.

meganrm:
> The module `rtree` is required to run `analyze`. Could we add it to the setup? Lmk if this is added already and I just have to re-run the setup.
> 
> image

It's not being imported by the code, so it might be a dependency of one of our required modules. Have you tried re-running the install script just to check? I haven't run into this

meganrm:
> Possible suggestion: could we copy the required `.obj` files to the output folder after packing? It would make it easier to drag and drop everything together onto the simularium viewer. Not sure how feasible this is for larger files.

Yeah that's a good idea

cellpack/autopack/Environment.py
```diff
@@ -1605,9 +1606,8 @@ def buildGrid(
             # save bb for current fill
             self.log.info("####BUILD GRID - step %r", self.smallestProteinSize)
             self.fillBB = boundingBox
-            self.grid = Grid(
-                boundingBox=boundingBox, space=self.smallestProteinSize, lookup=lookup
-            )
+            spacing = self.smallestProteinSize * 1.1547
```
mogres, line 1609: Just worked through the significance of 1.1547 = (2/sqrt(3)). This ensures that the smallest spherical protein completely encircles a cube made out of grid points with the given spacing, while touching all 8 corners. Mentioning it here for future reference.

meganrm: yeah, so it was how the grid was always supposed to be working, but if you look through the code it got removed and added all over the place... partially I think because of "smallestProteinSize" and "space" being used interchangeably. So I just do it once here when the "spacing" is being sent in

meganrm:
> * pyembree>=0.1.12

Thanks for running it. I'll knock the requirement down to this one

cellpack/test-recipes/test_compartment.json
```diff
@@ -12,17 +12,17 @@
...
         "runTimeDisplay": false,
         "overwritePlaceMethod": false,
-        "innerGridMethod": "jordan3",
+        "innerGridMethod": "raytrace",
         "boundingBox": [
             [
-                -600,
-                -1100,
-                -150
+                -0.600,
```
mogres, line 18: Is the bounding box recalculated during packing? This box looks pretty small compared to the compartments/ingredients in the recipe.

mogres:
Everything looks good to me! I was able to run the packing on Linux without any issues. It's great to have the functionality to pack meshes!

cellpack/test-recipes/test_compartment.json
meganrm: Yes it gets expanded to include the compartment, but the compartment is really small

PR #52, opened 2022-07-22
Feature/recipe loader

### LINKED ISSUE #67: Implement new recipe schema for spheres and sphere trees
## Use Case
The new schema has been designed and we need to be able to load the new files. Want to be able to load new recipe files with spheres and multi spheres 

## Solution
minimal changes to the internal workings of the algorithm, but keep all the loading logic in the recipe loader. 

## Alternatives

meganrm:
Problem
=======
closes #67

Solution
========
Loading the new recipe files with spheres and sphere trees. Greatly reduced the number of attributes for each ingredient class. 

## Type of change

Steps to Verify:
----------------
1. `pack -r 'test-recipes/v2/sphere_tree.json' -c 'packing-configs/debug.json'`
2. `pack -r 'test-recipes/v2/one_sphere.json' -c 'packing-configs/run.json'`

Thanks for contributing!


cellpack/autopack/Analysis.py
```diff
@@ -1093,7 +1095,7 @@ def plot_one_result_3D(self, filename, width=1000.0):
         for i in range(len(self.env.molecules)):
             m = self.env.molecules[i]
             pos.append(numpy.array(m[0]).tolist())
-            s.append(m[2].encapsulatingRadius ** 2)
+            s.append(m[2].encapsulating_radius ** 2)
```
meganrm, line 1098: Testing comment

cellpack/autopack/Environment.py
```diff
@@ -1158,21 +1018,45 @@ def setCompartmentMesh(self, compartment, ref_obj):
...
-        ingredient_type = arguments["Type"]
-        if ingredient_type == "SingleSphere":
-            ingr = SingleSphereIngr(**arguments)
-        elif ingredient_type == "MultiSphere":
+    def create_ingredient(self, recipe, **arguments):
+        ingredient_type = arguments["type"]
+
+        if ingredient_type == "single_sphere":
+            radius = arguments["radius"]
+            arguments = IOutils.IOingredientTool.clean_arguments(
```
meganrm, line 1026: this won't be needed once we have the migration code working, because things will either be in the new format, or converted to the new format

cellpack/autopack/Environment.py
```diff
@@ -477,6 +298,82 @@ def setSeed(self, seedNum):
         self.seed_set = True
         self.seed_used = SEED
 
+    def _prep_ingredient_info(self, composition_info, ingredient_name=None):
```
meganrm, line 301: @prismofeverything new code to look at

cellpack/autopack/Environment.py
```diff
@@ -66,8 +67,17 @@
...
+from cellpack.autopack.Grid import Grid
 from cellpack.autopack.MeshStore import MeshStore
+from cellpack.autopack.ingredient import Ingredient
 from cellpack.autopack.loaders.recipe_loader import RecipeLoader
+from cellpack.autopack.utils import (
+    cmp_to_key,
+    deep_merge,
+    ingredient_compare0,
+    ingredient_compare1,
+    ingredient_compare2,
```
prismofeverything, line 79: Is there a way to make the difference between these functions more clear in the name?

cellpack/autopack/Environment.py
```diff
@@ -321,15 +144,16 @@ def __init__(self, name="H", config=None, recipe=None):
         self.pickRandPt = not config["ordered_packing"]
         self.show_sphere_trees = config["show_sphere_trees"]
 
-        # TODO: this could come from recipe, the same way we're sending in config data
         self.boundingBox = numpy.array(recipe["bounding_box"])
+        self.name = name
```
prismofeverything, line 148: Any reason not to assign `self.name` immediately?

cellpack/autopack/Environment.py
```diff
@@ -477,6 +298,82 @@ def setSeed(self, seedNum):
...
+        for region_name, obj_keys in compartment_info.get(
+            "regions", {}
+        ).items():  # check if entry in compositions has regions
+            recipe = Recipe(name=f"{compartment_key}_{region_name}")
+            for key_or_dict in obj_keys:
+                if not isinstance(key_or_dict, dict):
+                    key = key_or_dict
+                    if key in self.compartment_keys:
+                        # deal with compartment
+                        return self._step_down(key)
```
prismofeverything, line 324: I think we can just call `_step_down` without returning, to allow for the rest of the internal compartments to process (may need to verify with debugger)

cellpack/autopack/Environment.py
```diff
@@ -477,6 +298,82 @@ def setSeed(self, seedNum):
...
+                        )
+                        self.create_ingredient(recipe, ingredient_info)
+                else:
+                    composition_info = key_or_dict
+                    ingredient_info = self._prep_ingredient_info(composition_info)
+                    self.create_ingredient(recipe, ingredient_info)
+            if region_name == "surface":
+                compartment.setSurfaceRecipe(recipe)
+            elif region_name == "interior":
+                compartment.setInnerRecipe(recipe)
```
prismofeverything, line 338: I know we want to support adding/defining custom regions, but we can do that in a subsequent PR.

cellpack/autopack/Environment.py
```diff
@@ -477,6 +298,82 @@ def setSeed(self, seedNum):
...
+                            composition_info = composition_dict[key]
+                            ingredient_info = self._prep_ingredient_info(
+                                composition_info, key
+                            )
+                            self.create_ingredient(external_recipe, **ingredient_info)
+                    else:
+                        composition_info = key_or_dict
+                        ingredient_info = self._prep_ingredient_info(composition_info)
+                        self.create_ingredient(external_recipe, **ingredient_info)
+            self.setExteriorRecipe(external_recipe)
```
prismofeverything, line 371: This code is similar to the definition of `_step_down`, can we have one function that covers both? Perhaps we can deal with the exterior recipe when `_step_down` is called from the root instead of inside this similar code, and only do all of this in one place?

prismofeverything: If that is substantial enough, we can save it for a future more focused PR to untangle the cytoplasm assumption

cellpack/autopack/ingredient/single_cube.py
```diff
@@ -12,118 +12,95 @@
...
-        self.bb = [-radii[0] / 2, radii[0] / 2]
-
-        self.positions = positions  # bottom left corner of cuboid
-        self.positions2 = positions2  # top right corner of cuboid
-        positions_ar = numpy.array(self.positions[0][0])
-        positions2_ar = numpy.array(self.positions2[0][0])
+        self.bb = bounds
 
+        self.lower_bound = bounds[0]  # bottom left corner of cuboid
+        self.upper_bound = bounds[1]  # top right corner of cuboid
```
prismofeverything, line 90: I think this leaves open the ability for `self.bb` to diverge/differ from `self.lower_bound/self.upper_bound`? Could be good to merge these into a single source of truth at some point.

cellpack/autopack/ingredient/single_sphere.py
```diff
@@ -20,113 +20,144 @@ class SingleSphereIngr(Ingredient):
 
     def __init__(
         self,
-        Type="SingleSphere",
+        radius,
+        type="single_sphere",
```
prismofeverything, line 24: Could we merge `SingleSphere` and `MultiSphere` some day? and treat a single sphere as a multisphere with a single sphere in it.... ?

cellpack/autopack/upy/simularium/simularium_helper.py
```diff
@@ -109,6 +109,16 @@ def __init__(self, master=None, vi=None):
...
 
+    @staticmethod
+    def format_rgb_color(color):
+        need_to_divide = False
+        for ele in color:
+            if ele > 1:
+                need_to_divide = True
+        if need_to_divide:
+            color = np.array(color) / 255
+        return matplotlib.colors.to_hex(color)
```
prismofeverything, line 120: This helps to deal with the two different representations of RGB colors (components with 0-255 vs 0.0-1.0) but leaves open a strange corner case if someone intends to have a color between 0-1 relative to a scale of 255, which in this case will become surprisingly white, instead of almost black.

prismofeverything: Perhaps we make this distinction based on the type of the value (float vs int) instead of the scale of the components?

cellpack/autopack/utils.py
```diff
@@ -1,5 +1,139 @@
+import collections
 import numpy
 
 
 def get_distance(pt1, pt2):
     return numpy.linalg.norm(pt2 - pt1)
+
+
+def ingredient_compare1(x, y):
+    """
```
prismofeverything, line 10: Can we rename these functions to reflect the actual difference? Looks like priority > 0 vs priority < 0

prismofeverything: Or perhaps combine all three into a single function that handles all cases?

prismofeverything: Pass in another argument that determines which regime we use to break ties based on priority?

prismofeverything (commented):
This is a monumental effort, comments below.

cellpack/autopack/Environment.py
meganrm: added an issue: https://github.com/mesoscope/cellpack/issues/68

cellpack/autopack/Environment.py
meganrm: same as above: https://github.com/mesoscope/cellpack/issues/68

mogres:
@meganrm could you please force-add the recipe for the sphere tree? Should be `/test-recipes/v2/sphere_tree.json`. Thanks!

meganrm:
> @meganrm could you please force-add the recipe for the sphere tree? Should be `/test-recipes/v2/sphere_tree.json`. Thanks!

I'm not sure what you mean, it's in this change set as a new file.

mogres:
> > @meganrm could you please force-add the recipe for the sphere tree? Should be `/test-recipes/v2/sphere_tree.json`. Thanks!
> 
> I'm not sure what you mean, it's in this change set as a new file.

I see it now! I got an error due to the leading `/`

PR #53, opened 2022-07-22
Feature/show sphere trees

meganrm:
Problem
=======
We want to be able to visualize the collision geometry 

Solution
========
Put spheres in the output file

## Type of change

Screenshots (optional):
-----------------------

-----------------------
1. main file/entry point
2. other important file

Thanks for contributing!


mogres (approved):
Looks good, thanks for adding this in!

PR #58, opened 2022-09-06
Feature/pdb offset

meganrm:
Problem
=======
The sphere files are centered, but the pdb files might not be. 

Solution
========
Added some scripts to get the offsets for the pdb files (can automate this in future iterations). 
 @mogres implemented an offset solution that Dan suggested to maintain the correct rotation

I hardcoded a rotation for one of the sphere trees, we can remove this once we have the sphere tree maker 

## Type of change

Screenshots (optional):
-----------------------
Screen Shot 2022-09-08 at 12 42 48 PM


mogres (approved):
Looks good to me. Thanks for adding this in!

PR #77, opened 2022-09-26
suggested changes to PR

meganrm:
Sorry this this #76 issue didn't explain enums well, but this is more what I was thinking. 

You did a really good job with the explanation you were given. The point of the enum is to replace the string values with a constant. so the code can tell you if you have mistyped something. IE, if you type `if type == "sing_sphere"` the code doesn't know that nothing is ever going to be called `sing_sphere`. But `type == INGREDIENT_TYPE.SINGLE_SPHERE` can't be mistyped. 

We want this typing for the new schema, the old schema isn't going to change, so the simple mapping like we did before is sufficient, but we're just mapping to enum values instead of strings.

### DIFF
--- a/cellpack/autopack/loaders/recipe_loader.py
+++ b/cellpack/autopack/loaders/recipe_loader.py
@@ -10,7 +10,7 @@ import cellpack.autopack as autopack
 from cellpack.autopack.loaders.util import create_file_info_object_from_full_path
 from cellpack.autopack.utils import deep_merge
 from .v1_v2_attribute_changes import (
-    ingredient_types,
+    ingredient_types_map,
     v1_to_v2_name_map,
     unused_attributes_list,
     convert_to_partners_map,
@@ -165,17 +165,10 @@ class RecipeLoader(object):
         new_ingredient = {}
         for attribute in list(old_ingredient):
             if attribute in v1_to_v2_name_map:
+                value = old_ingredient[attribute]
                 if attribute == "Type":
-                    converted_value = ingredient_types.convert(
-                        old_ingredient[attribute]
-                    )
-                    new_ingredient[
-                        v1_to_v2_name_map[attribute]
-                    ] = converted_value.get_key_name()
-                else:
-                    new_ingredient[v1_to_v2_name_map[attribute]] = old_ingredient[
-                        attribute
-                    ]
+                    value = ingredient_types_map[old_ingredient[attribute]]
+                new_ingredient[v1_to_v2_name_map[attribute]] = value
             elif attribute in unused_attributes_list:
                 del old_ingredient[attribute]
             elif attribute in convert_to_partners_map:
--- a/cellpack/autopack/loaders/recipe_loader.py
+++ b/cellpack/autopack/loaders/recipe_loader.py
@@ -165,8 +165,10 @@ class RecipeLoader(object):
         new_ingredient = {}
         for attribute in list(old_ingredient):
             if attribute in v1_to_v2_name_map:
-                value = old_ingredient[attribute]
                 if attribute == "Type":
+                    value = ingredient_types_map[old_ingredient[attribute]]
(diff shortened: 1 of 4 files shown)


cellpack/autopack/loaders/recipe_loader.py
```diff
@@ -165,17 +165,10 @@ def _migrate_ingredient(old_ingredient):
...
-                        old_ingredient[attribute]
-                    )
-                    new_ingredient[
-                        v1_to_v2_name_map[attribute]
-                    ] = converted_value.get_key_name()
-                else:
-                    new_ingredient[v1_to_v2_name_map[attribute]] = old_ingredient[
-                        attribute
-                    ]
+                    value = ingredient_types_map[old_ingredient[attribute]]
```
rugeli, line 170:
```suggestion
for attribute in list(old_ingredient):
            if attribute in v1_to_v2_name_map:
                if attribute == "Type":
                    value = ingredient_types_map[old_ingredient[attribute]]
                else: 
                    value = old_ingredient[attribute]
```

cellpack/autopack/loaders/recipe_loader.py
```diff
@@ -165,17 +165,10 @@ def _migrate_ingredient(old_ingredient):
...
+                new_ingredient[v1_to_v2_name_map[attribute]] = value
```
rugeli, line 171:
```suggestion
for attribute in list(old_ingredient):
            if attribute in v1_to_v2_name_map:
                if attribute == "Type":
                    value = ingredient_types_map[old_ingredient[attribute]]
                else: 
                    value = old_ingredient[attribute]
```

cellpack/autopack/loaders/recipe_loader.py
```diff
@@ -165,17 +165,10 @@ def _migrate_ingredient(old_ingredient):
         new_ingredient = {}
         for attribute in list(old_ingredient):
             if attribute in v1_to_v2_name_map:
+                value = old_ingredient[attribute]
                 if attribute == "Type":
```
rugeli, line 169:
```suggestion
for attribute in list(old_ingredient):
            if attribute in v1_to_v2_name_map:
                if attribute == "Type":
                    value = ingredient_types_map[old_ingredient[attribute]]
                else: 
                    value = old_ingredient[attribute]
```

rugeli: I'm wondering if we could bring line168 under an else statement so the value of "Type" won't need to be assigned twice?

cellpack/autopack/loaders/v1_v2_attribute_changes.py
```diff
@@ -43,19 +43,11 @@
...
-    def get_key_name(self):
-        return self.name.lower()
+ingredient_types_map = {
+    "SingleSphere": INGREDIENT_TYPE.SINGLE_SPHERE,
+    "MultiSphere": INGREDIENT_TYPE.MULTI_SPHERE,
+    "SingleCube": INGREDIENT_TYPE.SINGLE_CUBE,
+    "SingleCylinder": INGREDIENT_TYPE.SINGLE_CYLINDER,
+    "MultiCylinder": INGREDIENT_TYPE.MULTI_CYLINDER,
+    "Grow": INGREDIENT_TYPE.GROW,
+}
```
rugeli, line 53: great implementation! I see what you meant by "can't be mistyped" now

rugeli (approved):
The example in your comment made a clear sense on why we want to use enum. Hopefully I will be able to think about the purpose (or ask more clarify questions) on each task in further works. Thank you for your time explaining.

cellpack/autopack/loaders/recipe_loader.py
meganrm: yeah that's fine. I doesn't really matter either way

PR #89, opened 2022-10-07
changed format_version to one decimal place

### LINKED ISSUE #87: v2 recipes should be "2.0" not "2.0.0"
## Description
When we first started making v2 recipes we set the `format_version` as `2.0.0` but we decided we don't need 3 levels of precision for format version, but right now this is making the code fail. 

## Solution
All format_versions should only have two numbers, ie, `2.0`

rugeli:
Problem
=======
closes #87 

Solution
========
changed format_version to "2.0" and "1.0" from "2.0.0" and "1.0.0" 
tested recipes in simularium 

## Type of change

Screenshots (optional):
-----------------------
Screen Shot 2022-10-06 at 8 42 39 PM

### DIFF
--- a/cellpack/test-recipes/v2/example.json
+++ b/cellpack/test-recipes/v2/example.json
@@ -1,6 +1,6 @@
 {
   "version": "2.2.2",
-  "format_version": "2.0.0",
+  "format_version": "2.0",
   "name": "nested_compartment_test",
   "bounding_box": [[0, 0, 0], [100, 100, 100]],
   "gradients": {
@@ -45,8 +45,9 @@
       "perturb_axis_amplitude": 0.1,
       "encapsulating_radius": 100,
       "packing": {
-        "mode": "random" // instead of "packing_mode": "random",
+        "packing_mode": "random" // instead of "packing_mode": "random",
       },
+      // we decided to keep "packing_mode", 10/6/22 Ruge
       "principal_vector": [
         0,
         0,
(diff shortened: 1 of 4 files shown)


cellpack/test-recipes/v2/example.json
```diff
@@ -45,8 +45,9 @@
       "perturb_axis_amplitude": 0.1,
       "encapsulating_radius": 100,
       "packing": {
-        "mode": "random" // instead of "packing_mode": "random",
+        "packing_mode": "random" // instead of "packing_mode": "random",
```
meganrm, line 48: this should not be nested in `packing`. 
```
        "encapsulating_radius": 100,
        "packing_mode": "random" // instead of "packing_mode": "random",
```

cellpack/test-recipes/v2/nested.json
```diff
@@ -1,6 +1,6 @@
 {
-    "version": "1.0.0",
-    "format_version": "2.0.0",
+    "version": "1.0",
```
meganrm, line 2: versions here should still be 3 digits. we're just changing the `format_version`

cellpack/test-recipes/v2/one_sphere.json
```diff
@@ -1,5 +1,5 @@
 {
-    "version": "1.0.0",
+    "version": "1.0",
```
meganrm, line 2: again, this should still be 3 values

meganrm:
This can just be a PR to `main`. the staging branch is for thing that are a large block of changes that take multiple PRs

meganrm (approved):

PR #90, opened 2022-10-12
Feature/store json to v1/converted

### LINKED ISSUE #88: Save converted recipe as a `.json` in the `out` folder
## Use Case
It would be good to have old recipes converted into the new file format

## Solution

Write a function in the recipe loader that is called by `migrate`.

After conversion programmatically save the converted recipe_data as a `[RECIPE-NAME]-v2.json`

the `out` folder is defined in the config, for example: `https://github.com/mesoscope/cellpack/blob/main/packing-configs/run.json`

rugeli:
Problem
=======
closes #88 

Solution
========

- wrote a function `_save_converted_recipe` to save recipes to v1/converted as json files 
- added a flag argument in `__init__` in `recipe_loader`, `run.json` and in config data to avoid saving test recipes when we run pytest 
- ignored converted recipe data in `.gitignore`  

with @meganrm @mogres Thank you for your insights!

## Type of change


meganrm (approved):

cellpack/autopack/loaders/migrate_v1_to_v2.py
```diff
@@ -67,7 +67,7 @@ def migrate_ingredient(old_ingredient):
     for attribute in list(old_ingredient):
         if attribute in v1_to_v2_name_map:
             if attribute == "Type":
-                value = ingredient_types_map[old_ingredient[attribute]]
+                value = ingredient_types_map[old_ingredient[attribute]].value
```
meganrm, line 70: why did you make this change?

rugeli: Before making this change, I tried to call `_save_converted_recipe` in `_migrate_version`, it raised TypeError: Object of type INGREDIENT_TYPE is not JSON serializable. So we add `.value` to get the value str in group programming. Please let me know if there's better ways to do this, I can look into it.

meganrm: yeah, so when you save it, you can convert it to a string. But you shouldn't do that to all of the code, just in `_save_converted_recipe`

mogres: v2 recipes do not go through `migrate_ingredient` and have the ingredient type as string. This can cause issues while creating ingredients in the Environment (comparing enum to strings). We should think about converting everything to enum if we want to avoid using string representations. Thoughts?

meganrm: Yeah, we can convert them on load, which can be part of the check to make sure they're an allowed member. I'll make an issue.

meganrm: Although this check works: `INGREDIENT_TYPE.SINGLE_SPHERE == "single_sphere"`

meganrm: This is now fixed in a new PR

mogres (approved):
Looks good! It's nice to see the interface between the config loader and the recipe taking root.

PR #91, opened 2022-10-13
make sure there are not mistakes in format versions

meganrm:
Problem
=======
The migrate code can return an empty object if the format version is wrong 

Solution
========
1. Check to see if the format version is 2.0.0 instead of 2.0 and fix it. 
2. Raise an error if the format version is something other than 1.0 or 2.0 (for now, will change as we add more supported versions)

## Type of change

### DIFF
--- a/cellpack/autopack/loaders/recipe_loader.py
+++ b/cellpack/autopack/loaders/recipe_loader.py
@@ -115,11 +115,11 @@ class RecipeLoader(object):
             print("filename is None and not ingredient dictionary provided")
             return None
         return data
-    
+
     @staticmethod
     def _sanitize_format_version(recipe_data):
         if "format_version" not in recipe_data:
-            format_version = "1.0" # all recipes before we introduced versioning
+            format_version = "1.0"  # all recipes before we introduced versioning
         elif len(recipe_data["format_version"].split(".")) > 2:
             # We only use two places for format version, but people
             # might accidently include a third number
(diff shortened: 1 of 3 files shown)


cellpack/autopack/loaders/recipe_loader.py
```diff
@@ -116,10 +116,24 @@ def _request_sub_recipe(self, inode):
             return None
         return data
 
-    def _migrate_version(self, recipe, format_version="1.0"):
+    @staticmethod
+    def _sanitize_format_version(recipe_data):
+        if "format_version" not in recipe_data:
+            format_version = "1.0"  # all recipes before we introduced versioning
+        elif len(recipe_data["format_version"].split(".")) > 2:
```
mogres, line 123: maybe could add another check to account for something like `recipe_data["format_version"] = 2`? Not really necessary but covers more cases.

mogres (approved):
Looks good! Thanks for adding the tests :)

cellpack/autopack/loaders/recipe_loader.py
```diff
@@ -116,10 +116,30 @@ def _request_sub_recipe(self, inode):
...
+            # We only use two places for format version, but people
+            # might accidently include a third number
+            # ie 2.0.0 instead of 2.0
+            split_numbers = recipe_data["format_version"].split(".")
+            format_version = f"{split_numbers[0]}.{split_numbers[1]}"
+        elif len(recipe_data["format_version"].split(".")) == 1:
+            # We only use two places for format version, but people
+            # might accidently include a third number
+            # ie 2.0.0 instead of 2.0
```
prismofeverything, line 132: This comment is copied from the section above I believe? and does not apply to this section, which is about adding a point number if only a single number is provided.

cellpack/tests/test_recipe_loader.py
```diff
@@ -67,3 +67,40 @@ def test_resolve_objects():
...
+            "2.105",
+        ),
+        (
+            {"format_version": "2"},
+            "2.0",
+        ),
+        (
+            {"format_version": "1"},
+            "1.0",
+        ),
```
prismofeverything, line 102: These cover the cases we expect pretty well. Do we want to add some freaky cases like `"a;slfhaofhweihfkd@@@"` or `"(((((------.828282828.18181818"`? May not be necessary but you never know what people come up with.

prismofeverything (commented):
Good to see the version migration feature coming to life!

PR #94, opened 2022-10-14
fix ingredient type checks

### LINKED ISSUE #93: Convert "type" to enum on load
## Use Case
We want to be able to use enums instead of string representations in our code. 

## Solution
In the recipe loader, convert `type` to enum

## Alternatives

meganrm:
Problem
=======
closes #93 

Solution
========
We don't actually need to convert the type, we just need to make sure the enums are typed as strings so `ingredient_type == INGREDIENT_TYPE.SINGLE_SPHERE` correctly evaluates 

## Type of change


mogres (approved):
Looks good!

Side note: we should add the `collides_with_compartment` method for multi spheres at some point.

PR #96, opened 2022-10-14
Fix/clean up ingredient values

meganrm:
Problem
=======
We had old checks in the code from dealing with old recipes that are now being converted

Solution
========
This is mostly a PR that DRYs out the code. There is no functional changes. 

The main thing I wanted to do is get the Ingredient class based on the ingredient type, but that caused some circular import issues, so I cleaned up and moved around some functions so we have clearer dependancies. IE, Environment is no longer importing RecipeLoader. 

I also wrapped `deep_merge` so you can send in the key you're merging two objects on, and it deletes the key afterwards. Ie when we use "inherit" we don't want the resulting object to still have "inherit" after we've copied the inherited values in 

## Type of change

* Maintenance (non-breaking change which fixes an issue)


cellpack/autopack/Environment.py
```diff
@@ -151,9 +143,7 @@ def __init__(self, config=None, recipe=None):
 
         # saving/pickle option
         self.saveResult = "out" in config
-        self.out_folder = RecipeLoader.create_output_dir(
-            config["out"], name, config["place_method"]
-        )
+        self.out_folder = create_output_dir(config["out"], name, config["place_method"])
```
meganrm, line 146: I just moved this to utils because it was a static method anyway, and now we're not importing `RecipeLoader`, which seemed like a bad dependency for `Environment` to have

cellpack/autopack/ingredient/__init__.py
```diff
@@ -7,3 +8,17 @@
...
+    INGREDIENT_TYPE.MULTI_SPHERE: MultiSphereIngr,
+    INGREDIENT_TYPE.SINGLE_CUBE: SingleCubeIngr,
+    INGREDIENT_TYPE.SINGLE_CYLINDER: SingleCylinderIngr,
+    INGREDIENT_TYPE.MULTI_CYLINDER: MultiCylindersIngr,
+    INGREDIENT_TYPE.GROW: GrowIngredient,
+}
+
+
+def get_ingredient_class(ingredient_type):
+    return type_to_class_map[ingredient_type]
```
meganrm, line 24: I thought we might want to do other checks here, which is why I made this a function instead of just using the map directly

cellpack/autopack/loaders/recipe_loader.py
```diff
@@ -30,37 +29,10 @@ def __init__(self, input_file_path):
         autopack.current_recipe_path = os.path.dirname(self.file_path)
         self.recipe_data = self._read()
 
-    @staticmethod
-    def is_key(key_or_dict, composition_dict):
```
meganrm, line 34: these two methods have just been moved, not actually deleted

cellpack/autopack/Environment.py
```diff
@@ -999,51 +987,8 @@ def create_ingredient(self, recipe, arguments):
...
-            )
-            ingr = ActinIngredient(**arguments)
-        else:
-            radius = arguments["radius"]
-            arguments = IOutils.IOingredientTool.clean_arguments(
-                Ingredient.ARGUMENTS, **arguments
-            )
-            arguments["type"] = INGREDIENT_TYPE.SINGLE_SPHERE
-            ingr = SingleSphereIngr(radius, **arguments)
+        ingredient_class = ingredient.get_ingredient_class(ingredient_type)
```
mogres, line 990: This is great!

cellpack/autopack/Environment.py
```diff
@@ -999,51 +987,8 @@ def create_ingredient(self, recipe, arguments):
...
+        ingr = ingredient_class(**arguments)
```
mogres, line 991: Assuming the `clean_arguments` step happens in the ingredient classes?

mogres (approved):
Looks good to me!

cellpack/autopack/utils.py
```diff
@@ -137,3 +138,11 @@ def deep_merge(dct, merge_dct):
         else:
             dct[k] = merge_dct[k]
     return dct
+
+
+def expand_object_using_key(current_object, expand_on, lookup_dict):
+    object_key = current_object[expand_on]
+    base_object = lookup_dict[object_key]
+    new_object = deep_merge(copy.deepcopy(base_object), current_object)
+    del new_object[expand_on]
```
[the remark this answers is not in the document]
meganrm: we would have to include it as an attribute into the `Ingredient` class (and all inherited classes). I don't think it's really useful information after we've unpacked the data and loaded it

cellpack/autopack/Environment.py
meganrm: no, it's now that we're getting much more standard arguments from the loader so we don't need to do this extra step anymore. This was a stop gap while we were removing some old attributes in old recipes

PR #99, opened 2022-10-19
Feature/multi sphere

meganrm:
Problem
=======
We want to load format version 1.0 recipes with multi sphere ingredients

Solution
========
A few small fixes to make sure these load

## Type of change


mogres:
Is there a test recipe I could try to pack? Currently running into an error `AttributeError: 'MultiSphereIngr' object has no attribute 'collides_with_compartment'` while trying to pack `cellpack/test-recipes/sphere_tree.json`

cellpack/autopack/Compartment.py
```diff
@@ -1230,7 +1230,9 @@ def BuildGrid_trimesh(self, env, grdPos, vSurfaceArea, srfPts, idarray, mesh_sto
         mesh = mesh_store.get_mesh(self.gname)
         # voxelized
         self.log.info(f"{self.name}: CREATED MESH")
-        trimesh_grid_surface = creation.voxelize(mesh, pitch=env.grid.gridSpacing)
+        trimesh_grid_surface = creation.voxelize(
+            mesh, pitch=env.grid.gridSpacing / 2
+        ).hollow()
```
meganrm, line 1235: gets a better mesh voxelization especially with low grid spacing

cellpack/autopack/interface_objects/representations.py
```diff
@@ -119,7 +119,7 @@ def get_mesh_path(self):
         else:
             if self.mesh["path"] == "default":
                 return f"{self.DATABASE}/geometries/{self.mesh['name']}"
-            return f"{self.mesh['path']}{self.mesh['name']}"
+            return f"{self.mesh['path']}/{self.mesh['name']}"
```
meganrm, line 122: so we don't get concatenated paths

prismofeverything: Could be worth looking into using `pathlib.Path` for paths in general. They handle all the corner cases of concatenation and also help when using windows style paths. 

> from pathlib import Path
> Path('one/path') / 'further/paths'

prismofeverything (approved):
Looks good to me

PR #115, opened 2022-11-17
Feature/gradients

meganrm:
Problem
=======
#120 

Solution
========
Adjusted the old gradient system to work with our new setup 
Created new function that calculates the distance from a surface to the next bounding surface. 

## Type of change

Steps to Verify:
----------------
1. `pack -r cellpack/test-recipes/v2/peroxisomes_gradient.json -c packing-configs/run.json`

Screenshots (optional):
-----------------------
Screen Shot 2022-11-22 at 9 56 46 AM


cellpack/autopack/Environment.py
```diff
@@ -280,6 +280,23 @@ def _setup(self):
                 self.referenced_objects,
             ) = Recipe.resolve_composition(self.recipe_data)
             self.create_objects()
+        if self.use_gradient:
+            gradients = self.recipe_data["gradients"]
```
prismofeverything, line 284: Only mildly redundant to have a boolean `use_gradient` here along with the presence or absence of a gradient in `self.recipe_data["gradients"]`. Could we collapse this check into a check for existence in recipe_data?

cellpack/autopack/Environment.py
```diff
@@ -280,6 +280,23 @@ def _setup(self):
...
+                    )
+                self.set_gradient(gradient_name, gradient_data)
+
+    def get_compartment_object_by_name(self, compartment_name):
+        """
+        Returns compartment object by name
+        """
+        for compartment in self.compartments:
+            if compartment.name == compartment_name:
+                return compartment
```
prismofeverything, line 299: This is essentially reimplementing a dictionary lookup as a linear search.... I know we never have that many compartments but maybe we will some day? How plausible is it to just make `self.compartments` a dictionary? If you need the list you could just `self.compartments.values` etc.

cellpack/autopack/Environment.py
```diff
@@ -684,19 +698,16 @@ def sortIngredient(self, reset=False):
             if o.surfaceRecipe:
                 o.surfaceRecipe.sort()
 
-    def setGradient(self, **kw):
+    def set_gradient(self, name, gradient_data):
```
prismofeverything, line 701: Thanks for removing `**kw` here : )

cellpack/autopack/Gradient.py
```diff
@@ -73,7 +73,17 @@ def __init__(self, name, mode="X", description="", direction=None, bb=None, **kw
...
+            "X",
+            "Y",
+            "Z",
+            "-X",
+            "-Y",
+            "-Z",
+            "direction",
+            "radial",
+            "surface",
+        ]
```
prismofeverything, line 86: I wonder about these modes.... they could in general have associated parameters or something ("radius" for radial?). Could be worthwhile to turn these modes into objects with attributes at some point.

cellpack/autopack/Gradient.py
```diff
@@ -217,189 +228,121 @@ def pickPoint(self, listPts):
...
-        xl, yl, zl = bb[0]
-        xr, yr, zr = bb[1]
+        center = self.direction
+        max_distance = self.radius
+        distances = get_distances_from_point(master_grid_positions, center)
+        self.distances = (
+            numpy.where(distances < max_distance, distances, max_distance)
+            / max_distance
+        )
+        self.set_weights_by_mode()
```
prismofeverything, line 288: Great to see more numpy optimizations work their way in.

prismofeverything (approved):
Hey @meganrm looks great! Just a few comments but overall I think this is ready to go.

cellpack/autopack/Environment.py
```diff
@@ -2764,7 +2786,7 @@ def setupPanda(self):
             self.rb_panda = []
         for o in self.compartments:
             if o.rbnode is None:
-                o.rbnode = o.addShapeRB()  # addMeshRBOrganelle(o)
+                o.rbnode = o.addShapeRB(self)  # addMeshRBOrganelle(o)
```
mogres, line 2789: Could we update `o` to something more descriptive?

cellpack/autopack/Environment.py
meganrm: at the moment we're mostly always iterating over all compartments so I think it makes sense to have it be an array

PR #123, opened 2022-12-14
Calculate and scale distance between surfaces for gradients

### LINKED ISSUE #113: Recalculate surface distances while using gradients
## Use Case
While packing ingredients with gradients dependent on distance from surfaces, `surface_distances` is calculated when the grid is initialized. This calculation does not happen when the grid is loaded from a file, which makes re-running packings with previously computed grids take longer. Note that the surface distances may be different between packings depending on the position of the gradient generating object so it'd be ideal but time consuming to recalculate them for each packing.

## Solution
Create a separate function to calculate surface distances for all compartments and call it when weights for gradients are calculated. This function should be called after compartments are created and packed.

mogres:
Problem
=======
We want to be able to pack ingredients with a gradient that scales based on the distance between compartment surfaces. E.g. packing peroxisomes with a gradient going from the nuclear surface to the cell membrane.

Closes #113 , #115 

Solution
========
Updated the surface distance calculations to provide an option to scale to the next surface. These distances are now calculated using in-built trimesh functions.  

with @meganrm 

## Type of change

Change summary:
---------------
* Used in-built trimesh functions to calculate scaled distance between surfaces
* Created gradients scaled using inter-surface distances
* Added utilities to visualize this distance

Steps to Verify:
----------------
1. activate the virtual environment: `conda activate autopack`
2. run a packing of peroxisomes in cell with a gradient based on the distance between the nucleus and the cell membrane:
`pack -r cellpack/test-recipes/v2/peroxisomes_gradient.json -c packing-configs/sparse-packing.json`

Screenshots:
-----------------------
image
image

-----------------------
1. `cellpack/autopack/Compartment.py`

To Do:
1. This functionality calculates scaled surface distances for all nested compartments which may become time consuming. Might be useful to add an option to only calculate if a gradient uses these distances.


cellpack/autopack/Gradient.py
```diff
@@ -332,9 +337,12 @@ def build_axis_weight_map(self, bb, master_grid_positions, axis="X"):
...
         scaled_distances = self.distances
-        if max(scaled_distances) > 1.0:
-            self.log.error("MAX TOO BIG", max(scaled_distances))
-            # raise ValueError("distances have not been scaled to be from 0.0 to 1.0")
+
+        # TODO: change this back to check distances are assigned correctly
+        scaled_distances[scaled_distances > 1] = 1
+        scaled_distances[scaled_distances < 0] = 0
+        # if max(scaled_distances) > 1.0:
+        #     self.log.error("MAX TOO BIG", max(scaled_distances))
```
meganrm, line 345: why is this needed here? Shouldn't they already be scaled?

mogres: Good catch. Removed this and added an error check.

meganrm (approved):

PR #136, opened 2023-02-07
Adding recipe version to result file name

mogres:
Problem
=======
We want to differentiate between versions of the same recipe while keeping all the results in the same folder for analyses.

Solution
========
Append the recipe version to the name of saved files and data. 
~~NOTE: the recipe version is *NOT* appended to the grid file to avoid regenerating the grid when a recipe is changed.~~

## Type of change

Change summary:
---------------
* added recipe version to result file name
* added option to disable plotting during analyses

Steps to Verify:
----------------
1. Setup the cellPACK environment as shown in the README
1. run `pack -r cellpack/test-recipes/v2/peroxisomes_surface_gradient.json -c packing-configs/peroxisome_packing_config.json`
1. Change version in the recipe and re-run packing to verify results are saved with appended version names in the same folder

-----------------------
1. Environment.py
2. Analysis.py


mogres:
Unsure why linting fails. `black` doesn't raise any errors locally:
image

The same version of `black` is also installed during the check:
image

meganrm (approved):

PR #146, opened 2023-03-10
move files to make it clear what are examples and what are for tests

meganrm:
Problem
=======
We have some example recipes in this repo and then we have files that we use for tests/analysis. 

Solution
========
I moved everything we reference in tests or workflows into `cellpack/tests`. I moved everything we just have as examples (recipes and configs) into an `example` folder. 

## Type of change


mogres (approved):
This is very useful!

rugeli (approved):
Good idea to separate test recipes and examples!

Issue #139, opened 2023-02-08
Create sub directories in AWS per PR and delete them on merge


meganrm:
## Use Case
We want analysis results for each PR, but currently if we had two open PRs they will both be writing the same files to aws, so we could have a case were we're not seeing accurate results 

## Solution
*Please describe your ideal solution*
We could use the PR url to create a sub directory to store the specific results and then delete those files on merge 

## Alternatives
*Please describe any alternatives you've considered, even if you've dismissed them*

PR #150, opened 2023-03-28
Modified gh action scripts to create a new sub_directory in s3 whenev…

aditnath:
Problem
=======
When creating new PR's, the analysis pipeline was using the wrong analysis files when there were multiple PRs out at once. 
(https://github.com/mesoscope/cellpack/issues/139)

Solution
========
Talking with Saurabh, we decided that we would need to change 2 files:

analyze.yml: Makes a subdirectory in s3 for each new PR, and modifies the PR_analysis_config.yml file with the new path to the s3 sub_directory.
ci.yml: Cleanup step deletes the s3 sub_directory on merge 
Edit: Needed to create an entirely new file called cleanup.yml, since ci.yml runs on the creation of any PR, which is not the intended behavior. Cleanup.yml only deletes extraneous folders on closing of PRs.

with @mogres 

## Type of change


.github/workflows/analyze.yml
```diff
@@ -19,10 +19,22 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install .[all]
+    - name: Set PR name environment variable 
+      run: | 
+        echo "PR_NAME=$(( github.head_ref }}" >> $GITHUB_ENV
```
mogres, line 24: There's a mismatch here in the opening and closing brackets.

.github/workflows/analyze.yml
```diff
@@ -43,7 +55,7 @@ jobs:
     - name: Remove files in results bucket
       run: aws s3 rm  s3://cellpack-results --recursive
```
mogres, line 56: We should get rid of this step if we're taking care of this in cleanup

aditnath: yea good call

mogres:
@aditnath Looks like `analyze.yml` has an error:
image

I think a good proof of concept would be to have the analysis workflow run on this PR and have it post a comment on the thread.

.github/workflows/ci.yml
```diff
@@ -87,3 +87,20 @@ jobs:
       with:
         user: meganrm
         password: ${{ secrets.PYPI_TOKEN }}
+
+  cleanup:
```
meganrm, line 91: this workflow runs on every pull request, so wouldn't it create files and then immediately delete them? I think we want a separate workflow that has this job that is triggered by a pull request being merged. Have you tested this?

meganrm: I also looked at the output of this and $BRANCH_NAME is being interpreted as a string. You can set env variables directly, you don't need to run `echo "BRANCH_NAME=${{ github.ref }}" > $GITHUB_ENV`
https://docs.github.com/en/actions/learn-github-actions/variables

.github/workflows/analyze.yml
```diff
@@ -19,10 +19,22 @@ jobs:
       run: |
         python -m pip install --upgrade pip
         pip install .[all]
+    - name: Set PR name environment variable 
+      run: | 
+        echo "PR_NAME=${{ github.head_ref }}" >> $GITHUB_ENV
```
meganrm, line 24: same here, you can just define the env variable

.github/workflows/analyze.yml
```diff
@@ -19,10 +19,22 @@ jobs:
...
       run: |
         pack -r cellpack/tests/recipes/v2/test_spheres.json -c cellpack/tests/packing-configs/test_config.json
-        analyze -a cellpack/tests/analysis-configs/PR_analysis_config.json
+    - name: Modify JSON with PR sub_directory path
+      run: |
+        JSON=$(cat cellpack/tests/analysis-configs/PR_analysis_config.json)
+
+        NEW_JSON=$(echo "JSON" | jq '. + {"results_output_path": "https://cellpack-results.s3.us-west-2.amazonaws.com/${PR_NAME}/jitter"})
+
+        echo "$NEW_JSON" > cellpack/tests/analysis-configs/PR_analysis_config.json
```
meganrm, line 34: this is a cool way to solve this

.github/workflows/ci.yml
meganrm: If you forked this repo, you could be opening and closing your own PRs to make sure this got triggered correctly. Often when I'm doing GitHub actions I first try stuff out in my own dummy repo so I don't have to worry about reverting any changes.

meganrm: Also, once you fix the error in the main workflow, this'll run on this PR so you'll be able to see what's happening

meganrm: https://github.com/mesoscope/cellpack/actions/runs/4547324139

cellpack/tests/analysis-configs/PR_analysis_config.json
```diff
@@ -8,6 +8,6 @@
         "run_distance_analysis": true,
         "path_to_test_recipe": "cellpack/tests/recipes/v2/test_spheres.json",
         "path_to_results": "cellpack/tests/outputs/test_spheres/jitter",
-        "results_output_path": "https://cellpack-results.s3.us-west-2.amazonaws.com/jitter"
+        "results_output_path": "https://cellpack-results.s3.us-west-2.amazonaws.com/{branch_name}/jitter"
```
mogres, line 11: `jitter` should be replaced with `spheresSST` here and in analyze.yml

meganrm:
@aditnath can you convert this to a draft until you know it's working? that way we'll get an alert again when it's ready for review

mogres (approved):
Looks good to me! I can see the full analysis report comment with images on the PR conversation, and all the files in the s3 bucket. Just need to test the cleanup on merge part of the workflow, but we should be good to go after @meganrm has also taken a look.

.github/workflows/analyze.yml
```diff
@@ -22,7 +22,12 @@ jobs:
     - name: Pack test recipe
       run: |
         pack -r cellpack/tests/recipes/v2/test_spheres.json -c cellpack/tests/packing-configs/test_config.json
-        analyze -a cellpack/tests/analysis-configs/PR_analysis_config.json
+    - name: Modify JSON with PR sub_directory path
+      run: |
+        jq --arg branch_name "$GITHUB_REF_NAME" '.create_report.results_output_path |= "https://cellpack-results.s3.us-west-2.amazonaws.com/\($branch_name)/spheresSST"' cellpack/tests/analysis-configs/PR_analysis_config.json > cellpack/tests/analysis-configs/PR_analysis_config_temp.json
```
meganrm, line 27: this is still being interpreted as a string. (check the logs of the action that ran). I would use the GitHub object to get the branch name `https://docs.github.com/en/actions/learn-github-actions/contexts#github-context`. and make sure you're use ${{ }} when needed

.github/workflows/ci.yml
```diff
@@ -86,4 +86,4 @@ jobs:
       uses: pypa/gh-action-pypi-publish@master
       with:
         user: meganrm
-        password: ${{ secrets.PYPI_TOKEN }}
+        password: ${{ secrets.PYPI_TOKEN }}
```
meganrm, line 89: probably just revert this change

.github/workflows/cleanup.yml
```diff
@@ -0,0 +1,22 @@
...
+  cleanup:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Install AWS CLI
+      run: |
+        sudo apt-get update
+        sudo apt-get install -y awscli
+    - name: Remove files in results sub_dir
+      run: |
+        if aws s3 ls s3://cellpack-results/$GITHUB_REF/ ; then
```
meganrm, line 18: this variable is not being defined anywhere, and also probably interpreted as a string, so this won't work as is.

.github/workflows/analyze.yml
```diff
@@ -40,10 +45,8 @@ jobs:
         aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
         aws-region: us-west-2
     - uses: actions/download-artifact@v3
-    - name: Remove files in results bucket
-      run: aws s3 rm  s3://cellpack-results --recursive
     - name: Copy files to results bucket
-      run: aws s3 cp ./results s3://cellpack-results --recursive --acl public-read 
+      run: aws s3 cp ./results s3://cellpack-results/$GITHUB_REF_NAME/ --recursive --acl public-read
```
meganrm, line 49: Looks good, and looks like it's working but I don't get where GITHUB_REF_NAME is being defined

aditnath: From reading i think $GITHUB_REF_NAME is analagous to "{{ github.ref_name }}". Using it this way to make it more easily readable and able to be used with strings well

meganrm (approved):

PR #160, opened 2023-04-10
Live mode analysis

mogres:
Problem
=======
We want run analyses and create reports while packing a recipe.

Solution
========
added the option to specify analysis config file during packing. If the file is valid, an analysis workflow runs immediately after the packing without a separate invocation.

## Type of change

Change summary:
---------------
* update default values for analysis configs
* specify packing results path from command line instead of using a config
* save analysis results on class instance
* add docs
* update existing configs with new schema

Steps to Verify:
----------------
1. `conda activate autopack`
2. `pack -r cellpack/tests/recipes/v2/test_spheres.json -c cellpack/tests/packing-configs/test_config.json -a examples/analysis-configs/local_analysis_config.json` should pack the test recipe and create an analysis report at the specified location.

Keyfiles:
-----------------------
1. Analysis.py
3. pack.py
4. analyze.py


rugeli (approved):
Cool feature! I've compared each config file and learned which values we want to include in the analysis report, as well as the desired output location for the report and image. Looks good to me, thanks!

aditnath (approved):

PR #163, opened 2023-04-13
Init commit of MarkdownWriter class

aditnath:
Creating a generic markdown writer class in the cellpack writers repo

### DIFF
--- a/cellpack/autopack/writers/__init__.py
+++ b/cellpack/autopack/writers/__init__.py
@@ -76,7 +76,7 @@ class MarkdownWriter(object):
         self.report_md.create_md_file()
 
     # Image text must be a list, if list is not same length as list of filepaths, only 1st item in image_text is used
-    def add_images(self, header:str, image_text:typing.List[str], filepaths:typing.List):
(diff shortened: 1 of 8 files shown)


cellpack/autopack/writers/__init__.py
```diff
@@ -7,10 +7,99 @@
...
+            level=1,
+            title=header,
+            add_table_of_contents="n",
+        )
+        if len(image_text) == len(filepaths):
+            for i in range(len(filepaths)):
+                self.report_md.new_line(
+                    self.report_md.new_inline_image(
+                        text=image_text[i],
+                        path=str(self.output_path / filepaths[i])
```
mogres, line 90: We want this to be `output_image_location` and not `output_path`. The distinction is as follows:
1. `output_image_location` refers to the path used in the report to link to the image. This can be the AWS bucket path after the github workflow runs.
2. `output_path` is the path to which the markdown file itself is written. This is typically the root directory where packing outputs are stored.

mogres:
Looks like the tables are missing headers (check github bot comments on other PRs to identify missing headers)

rugeli:
Issued resolved in #276

Issue #165, opened 2023-04-14
Exporting tiff images from cellpack


mogres:
## Use Case
We would like to export packing images as `*.tiff` files for using them in other analyses. See [here](https://github.com/allen-cell-animated/feedback-research-repository/issues/88) for more details.

## Solution
Adding an option `export_output_as_tiff` to the packing configuration should create an output tiff files.

A possible approach could involve creating a voxelized grid with the voxel size as an input. Creating numpy arrays masked with positions of various ingredients, and using pre-existing libraries to output these arrays as tiff files.

Some packages that could be used to create tiff files from arrays:
1. `aicsimageio.writers` has the `OmeTiffWriter` class that can write numpy arrays to `ome.tiff` files
2. The `tifffile` library includes functionality to save arrays as tiff files

## Possible difficulties:
1. Voxelizing fiber ingredients can prove tricky
2. Voxelizing meshes

## proposed MVP
pack a sphere
convert the results to a multichannel tiff file
can view the resulting image in a program like FIJI

mogres:
Working example with a circular mask using `aicsimageio`:
```python
import numpy as np
from aicsimageio.writers import OmeTiffWriter

def create_circular_mask(x, y, z, center=None, radius=None):

    if center is None: # use the middle of the image
        center = (int(x/2), int(y/2), int(z/2))
    if radius is None: # use the smallest distance between the center and image walls
        radius = min(center[0], center[1], center[2], x-center[0], y-center[1], z-center[2])

    X, Y, Z = np.ogrid[:x, :y, :z]
    dist_from_center = np.sqrt((X - center[0])**2 + (Y-center[1])**2 + (Z-center[2])**2)

    mask = dist_from_center <= radius
    return mask

img = np.zeros((500,500,10))
img[create_circular_mask(*tmp.shape, [0, 0, 0], 200)] = 1

OmeTiffWriter.save(img.T, "output.ome.tif", dim_order="ZYX")
```

mogres:
Previous version of issue at #117

PR #166, opened 2023-04-15
Export packing outputs as a tiff file

mogres:
Problem
=======
See #165

Solution
========
1. The multi-channel image is represented as a numpy array with dimensions `C x X x Y x Z`. Each channel represents a single type of packed ingredient or compartment.
1. For `single_sphere` ingredients, the `ImageWriter` class creates a mask based on the position and radius of the sphere, and sets all pixels within this mask to 255. Other ingredient types are currently unsupported.
2. For compartments represented by meshes, the `ImageWriter` class creates a hollow or filled mask based on the provided options, and sets all pixels within this mask to 255.
3. We can also convolve the image with a point-spread function as needed. Currently available options are "gaussian" and "box".
4. We use the `aicsimagio` package to export the packing image as an `ome.tiff` file. The camera is set looking down the `-z` axis similar to a microscopy assay.

with @meganrm 

## Type of change

Change summary:
---------------
* added the `ImageWriter` class to export packing outputs as tiff images
* added options in the packing config to edit the output image properties, including:
    1.  the voxel size (in units of grid spacing) OR image size (voxels in each dimension)
    2. hollow or filled mesh representation

Steps to Verify:
----------------
1. `conda activate autopack`
5. `pack -r cellpack/tests/recipes/v2/test_image_export.json -c cellpack/tests/packing-configs/test_image_config.json`

Example packing of peroxisomes in the average nuclear and cell shape:
`pack -r examples/recipes/v2/peroxisomes.json -c cellpack/tests/packing-configs/test_image_config.json`

Screenshots:
-----------------------
image
image
image

Keyfiles:
-----------------------
1. `cellpack/autopack/writers/ImageWriter.py`


mogres:
- [ ] Include a setting to create hollow masks for spheres.
- [ ] Check with matheus about creating peroxisome meshes

meganrm:
It's hard to be sure, but looking at the screenshots of the magenta spheres, it seems like the image is flipped

mogres:
> It's hard to be sure, but looking at the screenshots of the magenta spheres, it seems like the image is flipped

Yeah, I changed the dimension order in a later commit. Let me update the screenshots

aditnath (approved):

rugeli (approved):
I installed aicsimageio and am able to see the figures and the tiff image. Not so sure about what circular mask is, but the changes look good to me!

cellpack/autopack/writers/ImageWriter.py
```diff
@@ -0,0 +1,256 @@
...
+from pathlib import Path
+import numpy
+from scipy.ndimage import convolve
+
+"""
+ImageWriter provides a class to export cellpack packings as tiff images
+"""
+
+
+class ImageWriter:
```
toloudis, line 11: I would suggest that this class should be called Voxelizer or Voxelization etc.  Writing the image out is only a tiny part of what is in here!

cellpack/autopack/writers/ImageWriter.py
```diff
@@ -0,0 +1,256 @@
...
+            image = numpy.transpose(image, axes=(3, 0, 2, 1))
+        return image
+
+    def create_voxelization(self):
+        """
+        Creates a voxelized representation of the current scene
+        """
+        channel_colors = []
+
+        for pos, rot, ingr, _ in self.env.molecules:
```
toloudis, line 180: these two for loops (over molecules and compartments) might be more properly class members of whatever type self.env is.  Then you could call self.env.collect_voxelizations(self.image_data) and the function would fill in the image_data object.

PR #172, opened 2023-05-26
Peroxisome gradient experiments: parallel packings, variable ingredient count

### LINKED ISSUE #137: Add ability to run packings in parallel on multiple CPUs
## Use Case
This would be useful while running multiple packings with the same recipe but different seeds for analysis.

## Solution
Use `concurrent.futures` to run the `doloop` function in parallel

### LINKED ISSUE #173: Pack ingredients with variable counts drawn from a distribution
## Use Case
While running multiple packings, we would like to add an option to vary the count of an ingredient based on a provided distribution.

E.g., for a sphere, the count can be drawn from a uniform distribution with a minimum of 5 and a maximum of 15.

Options can be specified in the composition for an object as follows:
```
"count_options": {
    "distribution": "uniform",
    "min": 5,
    "max": 15
}
```
We would like to have the option to use uniform, normal and specified discrete list to draw counts from.

## Solution
Before packing, run a function `update_count` in `environment` that resets `ingr.count` and `ingr.left_to_place` based on values drawn from the specified options.

### LINKED ISSUE #177: Packing recipes with mesh ingredients in parallel
## Problem
Packing a recipe with a mesh ingredient that has other ingredients in its interior does not work if `parallel: true`. This was partially fixed when `initialize_shape` was moved to later in the workflow but that caused packings from saved grids using `load_from_grid_file` to fail.

## Solution
A weird bug was observed during calculations of ray triangle intersections in `trimesh` where the bounds of the mesh ended up being very large values (~1E+308). After investigating, it was found that the bounds were being extracted from the `triangles_tree` object stored in the `trimesh` scene instance. 
The default behavior of `trimesh` is to cache these values on the mesh object as an `rtree` object. When the constructed grid is saved to a file using `pickle` these objects are expected to be saved in a binary format that can be loaded back from the file. However, due to unknown limitations in `pickle`, the `triangles_tree` object stored in the cache would not load properly and ended up being set to large values. 
One possible way to deal with this issue is to clear the cached value of `triangles_tree` while loading the grids from file. After implementing this workaround, parallel packings with meshes and ingredients in the interior worked as expected. The fix is implemented in #172.

mogres:
Description
=======
This is a collection of enhancements that enable running large scale experiments for peroxisome packings in cells.

Closes #173 
Closes #137 (partially)
Closes #177 

## Type of change

Change summary:
---------------
* refactoring analyses
* run packings in parallel (currently does not work with meshes)
* updates to ImageWriter - convolve with PSFs and create different projections
* pack ingredients with variable count
* power-law and exponential gradient weight modes
* inverting a gradient

Steps to Verify:
----------------
1. `conda activate autopack`
2. `pack -r cellpack/tests/recipes/v2/test_spheres.json -c examples/packing-configs/parallel_packing_config.json`

To test variable counts:
2. `pack -r cellpack/tests/recipes/v2/test_variable_count.json -c examples/packing-configs/parallel_packing_config.json`

Screenshots:
-----------------------
1. Running 100 iterations of the `test_spheres` recipe with `"parallel": true` takes ~1 minute
image

1. Running 100 iterations of the `test_spheres` recipe with `"parallel": false` takes ~10 minutes
image

Keyfiles:
-----------------------
Analysis.py


cellpack/autopack/Analysis.py
```diff
@@ -71,10 +75,13 @@ def __init__(
         else:
             self.packing_results_path = Path()
 
-        if self.env is not None:
+        if output_path is not None:
+            self.output_path = Path(output_path)
+        elif self.env is not None:
```
meganrm, line 80: is there a case were you have `output_path` and `self.env`?

cellpack/autopack/Analysis.py
```diff
@@ -678,10 +685,19 @@ def plot_occurence_distribution(self, ingr):
...
             occ,
-            self.figures_path / f"{ingr.name}_occurrence_{self.env.basename}.png",
+            self.figures_path
+            / f"{ingr.name}_occurrence_{self.env.basename}_lineplot.png",
             title_str=ingr.name,
             x_label="seed",
-            y_label="occurences",
+            y_label="occurrences",
+        )
+        self.histo(
```
meganrm, line 694: is there still a `self.histo`? It should be `self.env`

cellpack/autopack/Analysis.py
```diff
@@ -1328,35 +1356,71 @@ def calc_similarity_df(self, all_objs, ingredient_key, save_path=None):
...
+                        # similarity_score = (ad_stat.significance_level - 0.001) / (
+                        #     0.25 - 0.001
+                        # )
+                        # similarity_score = 1 - ad_stat.statistic
+
+                        # 2 sample ks
+                        ks_stat = stats.ks_2samp(arr1, arr2)
+                        similarity_score = 1 - ks_stat.statistic
+                        # similarity_score = ks_stat.pvalue
+
```
meganrm, line 1410: delete?

cellpack/autopack/Compartment.py
```diff
@@ -3120,6 +3119,76 @@ def getSurfaceInnerPoints(self, boundingBox, spacing, display=True, useFix=False
...
+            [
+                (X * voxel_size[0] - center[0]).flatten(),
+                (Y * voxel_size[1] - center[1]).flatten(),
+                (Z * voxel_size[2] - center[2]).flatten(),
+            ]
+        )
+
+        mesh = mesh_store.get_mesh(self.gname)
+
+        if hollow:
```
meganrm, line 3157: do you have a strong sense of the difference in these options? if so it would be good to include in the comments

cellpack/autopack/Environment.py
```diff
@@ -312,6 +340,35 @@ def setSeed(self, seedNum):
         self.seed_set = True
         self.seed_used = SEED
 
+    @staticmethod
+    def validate_ingredient_info(ingredient_info):
```
meganrm, line 344: this seems like it should be in ingredient?

cellpack/autopack/Environment.py
```diff
@@ -517,7 +576,7 @@ def save_result(
         self.saveGridLogsAsJson(self.result_file + "_grid-data.json")
         self.collectResultPerIngredient()
         self.store()
-        self.store_asTxt()
+        # self.store_asTxt()
```
meganrm, line 579: why commented out?

cellpack/autopack/Environment.py
```diff
@@ -999,8 +1100,9 @@ def restoreGridFromFile(self, gridFileName):
         aSurfaceGrids = []
         f = open(gridFileName, "rb")
         self.readArraysFromFile(f)
-        for compartment in self.compartments:
-            compartment.readGridFromFile(f)
+        for ct, compartment in enumerate(self.compartments):
+            # compartment.readGridFromFile(f)
```
meganrm, line 1104: delete?

cellpack/autopack/Environment.py
```diff
@@ -1334,32 +1412,29 @@ def buildGrid(self, rebuild=True):
             self.grid.filename = self.previous_grid_file
             if self.nFill == 0:  # first fill, after we can just reset
                 self.log.info("restore from file")
-                self.restoreGridFromFile(self.previous_grid_file)
+                # self.restoreGridFromFile(self.previous_grid_file)
+                self.restore_grids_from_pickle(self.previous_grid_file)
```
meganrm, line 1416: delete

cellpack/autopack/Environment.py
mogres: Added an option to save this

cellpack/autopack/Environment.py
mogres: Verified the saving and loading, the updated functions should work now

cellpack/autopack/Analysis.py
mogres: The function name has been updated

cellpack/autopack/Analysis.py
```diff
@@ -1295,6 +1311,18 @@ def run_analysis_workflow(
...
 
+    def normalize_similarity_df(self, similarity_df):
+        """
+        Normalizes the similarity dataframe
+        """
+        dims_to_normalize = self.get_list_of_dims() + ["pairwise_distance"]
+        for dim in dims_to_normalize:
+            values = similarity_df.loc[:, dim].values
+            similarity_df.loc[:, dim] = (values - numpy.min(values)) / (
+                numpy.max(values) - numpy.min(values)
```
rugeli, line 1322: will there be case like: numpy.max(values) - numpy.min(values) == 0? should we check to avoid division by zero errors?

cellpack/autopack/Analysis.py
```diff
@@ -2137,18 +2193,194 @@ def combine_results_from_ingredients(self, input_dict):
...
+        ingredient_angle_dict=None,
+        ingredient_occurence_dict=None,
+        ingredient_key_dict=None,
+        get_distance_distribution=False,
+        image_export_options=None,
+        show_grid=False,
+        plot_figures=False,
+    ):
+        """
+        Packs one seed of a recipe and returns the recipe object
```
rugeli, line 2214: Thank you for writing docstrings and comments!

rugeli (approved):

cellpack/autopack/Analysis.py
mogres: Good point! I can add a check for this

PR #170, opened 2023-05-05
Feature/firebase upload gradients

rugeli:
Problem
=======
What is the problem this work solves, including
completes gradient collection upload in #124 
closes #129

Solution
========
What I/we did to solve this problem
- created `GradientDoc` class to handle gradient data uploads if it is included in the recipe we are uploading to firebase 
- wrote tests for the gradient upload feature

## Type of change

* New feature (non-breaking change which adds functionality)

Change summary:
---------------
* recipes I ran for testing: `peroxisomes_surface_gradient.json` and `gradients.json`

Steps to Verify:
----------------
1. `conda activate autopack`
2. `upload -r [relative-path-to-a-gradient-recipe] -c [your-firebase-cred-path]`
3. `pytest cellpack`

Screenshots (optional):
-----------------------
Screenshot 2023-05-05 at 4 09 58 PM

PR #174, opened 2023-06-02
Feature/resovle grads in comp

rugeli:
Problem
=======

To avoid submitting a large PR for downloading recipes to review, I separated the resolve gradient code to this branch from the`run-recipes-from-firebase` branch. This pr could be merged into #170.

In the current open pull request #170, we replace gradient data with a gradient path in objects once the gradient is uploaded. We simply compare the grad paths in db_data and local_data to determine whether the new doc should be written or not. 

However, during the downloading process, we need to get the full gradient data from composition in order to compile the recipe data for packing, and ideally we want to get all reference data (objects, gradients, compositions) in the same loop. 

Solution
========

**Resolve gradient data in objects recursively within compositions without breaking the `should_write` checks** 

## Type of change

Steps to Verify:
----------------
1. `conda activate autopack`
2. `upload -r [relative-path-to-a-gradient-recipe] -c [your-firebase-cred-path]`

-----------------------
1. `DBRecipeHandler.py`

### DIFF
--- a/cellpack/autopack/__init__.py
+++ b/cellpack/autopack/__init__.py
@@ -394,7 +394,7 @@ def load_file(filename, destination="", cache="geometries", force=None):
             db_doc, _ = db.get_doc_by_id(collection="recipes", id=recipe_id)
             db_handler = DBRecipeHandler(db)
             downloaded_recipe_data = db_handler.prep_db_doc_for_download(db_doc)
-            return downloaded_recipe_data
+            return downloaded_recipe_data, database_name
         else:
             local_file_path = get_local_file_location(
                 file_path, destination=destination, cache=cache, force=force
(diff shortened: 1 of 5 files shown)


cellpack/autopack/__init__.py
```diff
@@ -281,7 +281,8 @@ def is_remote_path(file_path):
     @param file_path: str
     """
     for ele in DATABASE_NAME:
-        return ele in file_path
+        if ele in file_path:
+            return ele in file_path
```
mogres, line 285: Could just return `True` instead of making another comparison
```suggestion
            return True
```

mogres (approved):
Thanks for separating into smaller, more manageable PRs (I should take notes 😅)

Issue #140, opened 2023-02-10
Make packing results available online for shareability


mogres:
## Use Case
It would be useful to have simularium packing results available online (on S3 etc.) in order to share URLs for visualization.

## Solution
Create a new public sub-bucket on AWS S3 to store simularium results and use the URL to load visualizations. 

## Alternatives
Dropbox, Drive, Firebase, etc. are alternative hosting services that could be used.

mogres:
> I updated the CORs rules on our bucket so the links work now. http://localhost:9001/viewer?trajUrl=https://cellpack-results.s3.us-west-2.amazonaws.com/jitter/seed_0_test_spheres_test_config_1.0.0_results.simularium

Confirmed this is working on simularium viewer. @meganrm are we good to close this issue or do we want to wait on reorganizing simularium files separately?

mogres:
I like the idea of adding the metadata, perhaps we could embed it within the Simularium file description as a first pass?

One idea we could implement to save space is clearing out old packing files periodically (every 30 days?) via github actions.

meganrm:
In Simularium helper:
1. add `upload_to_s3` which uploads the results to `cellpack-results` using the [s3 python sdk ](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/s3-uploading-files.html) and then saves metadata to the firebase under `results` collection. the metadata should have:
key: results_name
- user_name (not sure how to get this right now)
- datestamp (time and date that it was run, I think firebase automates this)
- the aws url where the results are stored 

3. add `open_in_simularium` which takes in the was url and makes a Simularium url and opens it. ex: `https://simularium.allencell.org/viewer?trajUrl=https://cellpack-results.s3.us-west-2.amazonaws.com/jitter/seed_0_test_spheres_test_config_1.0.0_results.simularium`

rugeli:
- [ ] how do we want to handle aws credentials? in github secrets or save in `.env`?
- [ ] where do we call functions to store results and `open_in_simularium`? 
- [ ] it is possible to have the same system username, but unlikely. Do we need to address it? 
- [ ] concern: following the changes, the `pack` command will be performing multiple tasks

mogres:
- [ ] We should consider deleting old result files (>6 months) to conserve space on s3. This could be a github action.

mogres:
Packed recipe results now have an option to be saved on s3 and viewed on simularium!

PR #181, opened 2023-07-24
Feature/upload to s3

rugeli:
Problem
=======
closing #140 

Solution
========
- In `simularium_helper.py`, added functionalities to store results to S3 - `cellpack-results/simularium` , upload metadata to firebase, and open packing results in simularium app.   

with @meganrm @mogres 

## Type of change

Steps to Verify:
----------------
1. pack a local recipe: `pack -r [path-to-a-local-recipe] -c [path-to-a-config-file]`
2. pack a remote recipe: pack -r `firebase:recipes/[firebase-recipe-id] -c [path-to-a-config-file]`(make sure you have the recipe living in your firebase before packing)

-----------------------
1.`simularium_helper.py`
2.`AWSHandler.py`

### DIFF
--- a/cellpack/autopack/DBHandler.py
+++ b/cellpack/autopack/DBHandler.py
@@ -4,6 +4,8 @@ from deepdiff import DeepDiff
 
 from cellpack.autopack.utils import deep_merge
 
+from google.cloud.exceptions import NotFound
+
 
 class DataDoc(object):
     def __init__(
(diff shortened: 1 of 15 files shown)


cellpack/autopack/DBHandler.py
```diff
@@ -289,18 +330,40 @@ def should_write(self, db):
...
+class DBHandler(object):
+    firebase_handler = None
+
     def __init__(self, db_handler):
         self.db = db_handler
+        if isinstance(db_handler, FirebaseHandler):
+            DBHandler.firebase_handler = db_handler
         self.objects_to_path_map = {}
         self.comp_to_path_map = {}
+        self.grad_to_path_map = {}
```
meganrm, line 366: This seems like changes that are outside of the scope of this PR?

rugeli:
This PR has separated into #189 #198 and #201

Issue #176, opened 2023-06-14
Check if PR exists before running Analysis github action


mogres:
## Use Case
Currently the comment workflow in the analysis github action fails if a correponding PR does not exist

## Solution
We want to check for the existence of a PR based on the commit sha or other environment variables before running the analysis action. If a PR does not exist, the action should not run.

mogres:
More details at #183

PR #184, opened 2023-07-27
Fix/update comment

rugeli:
Problem
=======
#176 

Solution
========
the workflow will on be triggered when:
1. a commit merged into `main`
2. a pull request has been opened/synchronized/reopened
Screenshot 2023-08-02 at 2 55 04 PM

## Type of change

Steps to Verify:
----------------
1. Push a commit on a branch that is not associated with a pull request---(the Analyze workflow will not be triggered)

### DIFF
--- a/.github/workflows/analyze.yml
+++ b/.github/workflows/analyze.yml
@@ -1,10 +1,28 @@
 name: Analyze
 
-on: push
+on: 
+  push:
+    branches:
+      - fix/update-comment
 
 jobs:
+    Check_pr:
+        runs-on: ubuntu-latest
+        outputs:
+            is_pr: ${{ steps.check.outputs.is_pr }}
+        steps:
+            - name: Check if PR
+              id: check
+              run: |
+                  if [[ "${{ github.event.pull_request }}" != "" ]]; then
+                    echo "::set-output name=is_pr::true"
+                  else
+                    echo "::set-output name=is_pr::false"
+                  fi
     Analyze:
         runs-on: ${{ matrix.os }}
+        needs: [Check_pr]
+        if: needs.Check_pr.outputs.is_pr == 'true'
         strategy:
             matrix:
                 python-version: [3.9]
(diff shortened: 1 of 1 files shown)


mogres:
This looks great! We could also skip the entire workflow (including the analysis part) if a PR does not exist. What do you think about moving the check further up in the `.yml` file? This will save us storage and processing time. Or do we want to keep the results on AWS as a test?

image

rugeli:
> This looks great! We could also skip the entire workflow (including the analysis part) if a PR does not exist. What do you think about moving the check further up in the `.yml` file? This will save us storage and processing time. Or do we want to keep the results on AWS as a test?
> 
> image

Good suggestion! If we don't have a strong need for keeping results from non-PR commits in AWS, then I'll go for saving storage and processing time.

.github/workflows/analyze.yml
```diff
@@ -3,8 +3,24 @@ name: Analyze
...
 jobs:
+    Check_pr:
+        runs-on: ubuntu-latest
+        outputs:
+            is_pr: ${{ steps.check.outputs.is_pr }}
+        steps:
+            - name: Check if PR
+              id: check
+              run: |
+                  if [[ "${{ github.event.pull_request }}" != "" ]]; then
```
mogres, line 14: Looks like `github.event.pull_request` is empty for this PR, which results in this statement evaluating to `False`. Maybe test it by checking what is the value stored in `github.event.pull_request`?

mogres:
Ran the following tests:

1. Pushing a commit to a draft PR triggers the workflow
2. Pushing a commit to a non-PR branch does not trigger the workflow
3. Pushing a commit to an existing PR triggers the workflow
4. Pushing multiple commits updates the existing comment

This is expected behavior and we should be good to merge!

mogres (approved):

meganrm (approved):
Looks good, wait for #185 to be merged.

PR #191, opened 2023-09-08
add upload_results as a config option

### LINKED ISSUE #190: make posting results to s3 a config option
## Use Case
if we run 100s of packings we probably don't want to open 100s of browser windows 

## Solution
in the config file add a `upload_results` that defaults to `true` 

## Alternatives

rugeli:
Problem
=======
closes #190 

Solution
========
- added a key `upload_results in the config file
- check config file settings before calling `post_and_open_file`, to prevent excessive uploads to S3 as well as to avoid opening multiple browser tabs unnecessarily 

the action takes place when:
- If `upload_results` is present in config data, it uses whatever value is specified: this allows the user to disable the opening of tabs automatically if they wish. 
Or
- If `upload_results` is missing in config data, we check `number_of_packings` and only upload/post if it is less than 1.

with @mogres   

## Type of change

Steps to Verify:
----------------
1. Pack a local recipe with or without the "upload_results" option in the config file


cellpack/autopack/writers/__init__.py
```diff
@@ -175,7 +175,8 @@ def save_as_simularium(self, env, all_ingr_as_array, compartments):
         file_name = env.helper.writeToFile(
             env.result_file, env.boundingBox, env.name, env.version
         )
-        autopack.helper.post_and_open_file(file_name)
+        if env.config_data is None or env.config_data.get("upload_results", True):
```
mogres, line 178: This check is a bit confusing for me. We're uploading and posting results if the `config_data` is missing or if `upload_results` is not set?

I would consider adding another check on `number_of_packings` to make sure we don't end up opening multiple browser tabs if `upload_results` is not set or missing on a config with 100s of packings.
```suggestion
        if env.config_data is None or env.config_data.get("upload_results", env.config_data.get("number_of_packings", 1) <= 1):
```

rugeli: Thank you for your suggestion of adding a check for `number_of_packings`! If `number_of_packings` is always going to be specified in parallel packings, then we should be able to prevent opening multiple browser tabs by just checking the `number_of_packings` key.

That way, we wouldn't need to worry about adding the `upload_results` key to the config.

For example, we could use `if env.config_data.get("number_of_packings", 1) <= 1):`

cellpack/autopack/writers/__init__.py
```diff
@@ -175,7 +175,8 @@ def save_as_simularium(self, env, all_ingr_as_array, compartments):
         file_name = env.helper.writeToFile(
             env.result_file, env.boundingBox, env.name, env.version
         )
-        autopack.helper.post_and_open_file(file_name)
+        if env.config_data.get("number_of_packings", 1) <= 1:
```
mogres, line 178: Thanks for incorporating this! However, I think keeping the `upload_results` option is still a good idea and we should only fall back on `number_of_packings` check if `upload_results` is missing from the config. 

My suggestion was:
`if env.config_data is None or env.config_data.get("upload_results", env.config_data.get("number_of_packings", 1) <= 1):`
which only uploads results and opens a tab if following conditions are met:
1) Config data is missing
Or
2) If `upload_results` is present in config data, it uses whatever value is specified: this allows the user to disable the opening of tabs automatically if they wish. Having this option would be useful in my opinion.
Or
3) If `upload_results` is missing in config data, we check `number_of_packings` and only upload/post if it is less than 1.
Let me know if this makes sense.

cellpack/autopack/writers/__init__.py
```diff
@@ -175,7 +175,10 @@ def save_as_simularium(self, env, all_ingr_as_array, compartments):
         file_name = env.helper.writeToFile(
             env.result_file, env.boundingBox, env.name, env.version
         )
-        autopack.helper.post_and_open_file(file_name)
+        if env.config_data is None or env.config_data.get(
+            "upload_results", env.config_data.get("number_of_packings", 1) <= 1
+        ):
```
meganrm, line 180: two things: I don't thin env.config_data can ever be none, so I don't think you need that check
I think you could do 3 or less. I have run some multiple packings recently that I would like to see the results of.

examples/packing-configs/peroxisome_packing_config.json
```diff
@@ -14,5 +14,6 @@
     "spacing": 2.5,
     "use_periodicity": false,
     "show_sphere_trees": false,
-    "load_from_grid_file": true
+    "load_from_grid_file": true,
+    "upload_results": false
```
meganrm, line 18: all config settings should also have a default setting in the config loader

mogres (approved):
Looks good to me! I think we can build upon this by using a single simularium file with multiple packings using `time` as an index.

Should be good to merge once conflicts are resolved!

PR #192, opened 2023-09-18
Fix assignment of compartment grid points

mogres:
Problem
=======
This PR includes various updates to grid creation and visualization and fixes errors in `main` while trying to pack nested compartments.

When assigning whether grid points are inside or outside a compartment, only grid points with a compartment id of 0 were checked and updated. This caused an issue if the compartments were nested, and the outer compartment was processed before the inner (as it should). 

Solution
========
`points_in_mesh` now updates the identity of grid points where compartment id <= the id of the current compartment being assigned. This assumes that larger compartments are created before smaller ones.

The `buildgrid_trimesh` function and distance calculations in `gradient.py` was updated to streamline and speed up grid creation. Visualization of grid points was cleaned up to remove unused values, improve readability and visibility.

Change summary:
---------------
Gradient calculations:
* removed normalization of surface distances while setting gradient attributes, this is now done while calculating the weights
* ignore nan values while scaling distances
* avoid calculating surface distances if not needed for a gradient. This calculation is computationally expensive and only needs to be performed if the gradient mode is `surface`

Grid creation:
* reduced chunk size for determining grid points inside-outside compartments
* **allow reassigning grid points for downstream compartments**: this assumes that outer compartments are created before the inner nested compartments. 

Grid visualization:
* sort grid point values and remove nans before visualizing
* set the size of grid points relative to the grid spacing

Steps to Verify:
----------------
`pack -r examples/recipes/v2/peroxisomes_surface_gradient.json -c examples/packing-configs/peroxisome_packing_config.json`

Screenshots (optional):
-----------------------
Time for packing `peroxisomes_surface_gradient`:

`fix/multiple_compartments`: image

-----------------------
1. Compartment.py
2. Gradient.py

### DIFF
--- a/cellpack/autopack/Compartment.py
+++ b/cellpack/autopack/Compartment.py
@@ -1326,7 +1326,9 @@ class Compartment(CompartmentList):
 
         point_compartment_ids = compartment_ids[points_in_encap_sphere]
         # largest compartments need to be created first for this to work
-        point_ids_to_assign = points_in_encap_sphere[numpy.abs(point_compartment_ids) < number]
+        point_ids_to_assign = points_in_encap_sphere[
+            numpy.abs(point_compartment_ids) < number
+        ]
         point_positions = numpy.float16(master_grid_positions[point_ids_to_assign])
 
         # check surface points
--- a/cellpack/autopack/Compartment.py
+++ b/cellpack/autopack/Compartment.py
@@ -1077,7 +1077,9 @@ class Compartment(CompartmentList):
             len(env.grid.masterGridPositions),
         )
 
-    def set_surface_distances(self, env, master_grid_positions):
+    def set_surface_distances(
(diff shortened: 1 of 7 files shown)


meganrm (approved):
Looks good to me

rugeli:
Thanks for fixing it! I can finally run and pack the `peroxisomes_surface_gradient` recipe. However, it took me 1212.10s. Should we look into it further? 
 
Screenshot 2023-09-22 at 1 44 02 PM

mogres:
> Thanks for fixing it! I can finally run and pack the `peroxisomes_surface_gradient` recipe. However, it took me 1212.10s. Should we look into it further?
> 
> Screenshot 2023-09-22 at 1 44 02 PM

Did you run this on your local machine? If so, the timing sounds about right! I was running this on the dgx workstation which would be faster than a laptop.

rugeli (approved):

PR #196, opened 2023-09-26
Fix gradient inversion

mogres:
Problem
=======
Gradient inversion used distances instead of weights. This could result in unexpected inversion.

Solution
========
The type of inversion can be specified as an option now. `invert: "distance"` inverts the scaled distances and `invert: "weight"` inverts the weights after calculation.

## Type of change

Change summary:
---------------
* add option to specify inversion using weights or distances
* add default values
* update test recipes to visualize inversion
* add credentials file to gitignore

Steps to Verify:
----------------
1. `pack -r cellpack/tests/recipes/v2/test_mesh_image_gradient.json -c cellpack/tests/packing-configs/test_mesh_config.json`
2. Change `invert: "weight"` to `invert: "distance"` to visualize alternative inversion.

Screenshots:
-----------------------
Inverting distances: 
image

Inverting weights:
image

Keyfiles:
-----------------------
1. Gradient.py

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -120,3 +120,6 @@ cellpack/tests/outputs/
 # data
 data/
 **/converted/*
+
+# credentials
+.creds
--- a/cellpack/autopack/Gradient.py
+++ b/cellpack/autopack/Gradient.py
@@ -207,14 +207,16 @@ class Gradient:
 
         self.scaled_distances = self.get_normalized_values(self.distances)
 
-        if self.invert == "distance":
-            self.scaled_distances = 1.0 - self.scaled_distances
-
-        if (max(self.scaled_distances) > 1.0) or (min(self.scaled_distances) < 0.0):
+        if (numpy.nanmax(self.scaled_distances) > 1.0) or (
+            numpy.nanmin(self.scaled_distances) < 0.0
+        ):
             raise ValueError(
                 "CHECK CALCULATED DISTANCES",
                 f"Max: {max(self.scaled_distances)}, Min: {min(self.scaled_distances)}",
             )
+
+        if self.invert == "distance":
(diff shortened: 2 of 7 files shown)


rugeli:
Looks good to me! I was able to pack the test recipe, however, it's taking a very long time to load the model in Simularium, I'm still waiting to visualize the result. 

One question(might not related to this PR):
In terminal logs, the message `Packing test_mesh_image_gradient_1.0.0:` appeared twice with the first one remaining at 0% progress. Should we make any modifications to address this if one of them is being skipped or something?
Screenshot 2023-09-27 at 2 27 56 PM

mogres:
@rugeli,  were you able to visualize the output? It might help if you change the radius of the `primitive_sphere` to `1` so that there are fewer grid points to visualize.

As for the progress bar, it gets reprinted/interrupted when something else gets printed to the terminal. The `Known pipe types` message seems to have reset the bar -- we can look into what causes that message and whether we can prevent it from being displayed.

rugeli:
> @rugeli, were you able to visualize the output? It might help if you change the radius of the `primitive_sphere` to `1` so that there are fewer grid points to visualize.
> 
> As for the progress bar, it gets reprinted/interrupted when something else gets printed to the terminal. The `Known pipe types` message seems to have reset the bar -- we can look into what causes that message and whether we can prevent it from being displayed.

Thanks for getting back to me! I wasn't able to visualize the output yesterday. I just changed the `primitive_sphere` to `1`, but the simularium has been in loading for the past 20 minutes.

mogres:
@rugeli the recipe has been updated so you should be able to run it faster and visualize in simularium. Let me know if there are still any issues

meganrm (approved):

PR #198, opened 2023-09-27
Feature/run inherited objects

### LINKED ISSUE #171: run firebase recipes that have "inherit" in objects
## Use Case

In firebase database,  object docs don't contain the key `inherit` when we upload them, but we need to find a way to revert inherit object (e.g `base`) into `objects` in the ready-for-packing recipe from the requested db recipe

## Solution
- when uploading the full recipe data, any existing `inherit` key should remain present
- if the `inherit` key is in object data, the corresponding inherited object name should be replaced with a reference path

## Alternatives

rugeli:
Problem
=======
closes #171 

Solution
========

- added a flag `resolve_inheritance` within the`_read()` method of `recipe_loader.py`
- when getting the full recipe data for upload, set the flag to false so the inherited objects won't be resolved, and the key `inherit` will be retained
- When uploading an individual object and an "inherit" key is identified, the reference of the inherited object will be linked
- resolved conflicts and bugs that arose from the inheritance reconstruction during uploading and downloading 

## Type of change

Steps to Verify:
----------------
1. `upload -r examples/recipes/v2/gradients.json`
2. `pack -r firebase:recipes/gradients_v-default`

Screenshots (optional):
-----------------------
Before - the inheritance key was resolved when receiving the full recipe data for uploading 
Screenshot 2023-09-27 at 1 30 24 PM
After - the key `inherit` is now retained and its value points to a reference   
Screenshot 2023-09-27 at 1 13 27 PM

-----------------------
1. `DBRecipeHandler.py`
3. `recipe_loader.py`

### DIFF
--- a/cellpack/autopack/DBRecipeHandler.py
+++ b/cellpack/autopack/DBRecipeHandler.py
@@ -336,6 +336,12 @@ class ObjectDoc(DataDoc):
                 ] = ObjectDoc.convert_positions_in_representation(position_value)
         return convert_doc
 
+    @staticmethod
+    def _object_contains_grad_or_inherit(obj_data):
+        return (
+            "gradient" in obj_data and isinstance(obj_data["gradient"], dict)
+        ) or "inherit" in obj_data
+
     def should_write(self, db):
         docs = db.get_doc_by_name("objects", self.name)
         if docs and len(docs) >= 1:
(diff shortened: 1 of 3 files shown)


mogres (approved):
Looks good to me! Test recipe upload and packing ran as expected.

meganrm (approved):

PR #203, opened 2023-10-11
fix/S3 uploads and new browser tabs handling 

rugeli:
Problem
=======
our current conditional statement ` if env.config_data.get("upload_results", env.config_data.get("number_of_packings", 1) <= 1):` on `main` prioritizes checking `upload_results` with its default value set to `False`. So the check will be skipped if `upload_results` is not explicitly set to `True` in the config file

Solution
========
changed the if-statement to evaluate `upload_results` and `number_of_packings` individually

with @meganrm 

## Type of change


mogres (approved):
Looks good to me! Having the `upload_results` set to `False` by default would not have worked with the previous check.

Issue #209, opened 2023-10-27
Parallel packing doesn't work with store multiple results in one simularium file


meganrm:
## Description
if you run a multiple packing config with parallel on, you get error like "AttributeError: 'Environment' object has no attribute 'result_file'" because the env has been cleared out or doesn't quite exist across the different threads. 

## Expected Behavior
running with parallel=true should work the same as with parallel=false

## Reproduction
*A minimal example that exhibits the behavior.*

``` 
pack -r 'cellpack/examples/recipes/v2/one_sphere.json' -c 'cellpack/cellpack/tests/packing-configs/test_parallel_config.json'
```

## Environment
*Any additional information about your environment*

mogres:
This may be related to environment attributes not updating when run in parallel. Setting `show_grid_plot` to `False` may prevent errors but will need to be tested.

mogres:
Running a test recipe:
```
pack -r examples/recipes/v2/one_sphere.json -c cellpack/tests/packing-configs/test_parallel_config.json
```
shows that the error occurs in line 2490 of `Analysis.py`. Commenting this out creates the individual simularium files but not the aggregate one.

Setting `"parallel": false` in the config file creates the aggregate simularium file. Weirdly enough, the packed sphere has a different color in the last time step for some reason. ([Link to simularium file](https://simularium.allencell.org/viewer?trajUrl=https://cellpack-results.s3.us-west-2.amazonaws.com/simularium/results_one_sphere_test_parallel_config_1.0.0_all.simularium))

PR #210, opened 2023-10-27
Feature/store runs in time

### LINKED ISSUE #36: Use "time" in simularium files to compare different packing runs
## Use Case

Ran multiple packing runs on the same recipe maybe with different inputs or just different seeds and want to compare them 

## Solution

Store a simularium file that has each packing result as a different time point

## Alternatives

meganrm:
Problem
=======
closes #36 

Solution
========
created a mapping from seed number to results to store all the objects in one simularium file 
with @mogres @kvegesna @rugeli 

## Type of change

Steps to Verify:
----------------
1. change the parallel setting to false in the config file 
2. ` pack -r 'cellpack/examples/recipes/v2/one_sphere.json' -c '/cellpack/cellpack/tests/packing-configs/test_parallel_config.json'`
3. should open the _all file in simularium with multiple packings


meganrm (commented):
I can't approve this because I opened the PR, but thanks for the more recent changes Saurabh, they look good to me

mogres (approved):
#209 has details about a bug while running this in parallel. Other than that, this looks good to go. Very useful update!!

mogres (changes requested):
~~I think we should bring back the check for  `upload_results` option before merging this.~~

Just realized I added this in earlier 😅

mogres (approved):

rugeli (approved):

rugeli:
Thank you for adding this feature and the config checks, looks good to me! #209 seems like a good topic for our next group programming session

PR #211, opened 2023-10-30
Maintenance/update readme

rugeli:
Problem
=======
In README, the current paths for example pack recipes need to be updated
the current link to README AWS S3 section in `simularium_helper` results in a 404

Solution
========
updated paths and link

## Type of change

Steps to Verify:
----------------
1. pack the example recipes in README
2. click the link to README's AWS section in `simularium_helper`

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,8 +15,9 @@ algorithm to pack molecular recipes
 3. `pip install -e .[dev]`
 
 ### Run pack code
-1. example pack recipe : `pack -r examples/recipes/v1/NM_Analysis_FigureB1.0.json -c examples/packing-configs/run.json`
-2. example pack from remote : `pack -r  github:recipes/NM_Analysis_FigureB1.0.json  -c examples/packing-configs/run.json`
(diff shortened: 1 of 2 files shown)


meganrm (approved):

mogres:
Could also add a `v2` recipe as an example in the readme.

rugeli:
I have added a v2 nested recipe as an example. would another v2 recipe be more suitable for readme? Additionally, I have noticed the v2 sphere tree packing seems off (`pack -r examples/recipes/v2/sphere_tree.json -c examples/packing-configs/debug.json`), we might want to address that in a separate pr.

rugeli:
> I have added a v2 nested recipe as an example. would another v2 recipe be more suitable for readme? Additionally, I have noticed the v2 sphere tree packing seems off (`pack -r examples/recipes/v2/sphere_tree.json -c examples/packing-configs/debug.json`), we might want to address that in a separate pr.

we will use V2 one_sphere recipe as an example.

mogres (approved):
looks good, thanks for updating!

PR #215, opened 2023-12-12
Updates for variable count and size experiments

mogres:
Problem
=======
This PR contains updates related to running packing experiments with variable count and size distributions.

Solution
========
These updates include changes to grid file loading and setting random seeds for packing replicates.

## Type of change

Change summary:
---------------
* Allow specifying a pre-existing grid file to load from. If the specified file does not exist, it will be created at that location.
* Updates to match the number of random seeds with the number of packing replicates. If `number_of_packings > 1`, a list of random seeds is generated to match. The random seeds can also be explicitly specified in the recipe itself.

Steps to Verify:
----------------
1. Run `pack -r cellpack/tests/recipes/v2/test_variable_count.json -c cellpack/tests/packing-configs/test_variable_count_config.json` in an activated conda environment


cellpack/autopack/utils.py
```diff
@@ -257,3 +257,26 @@ def get_value_from_distribution(distribution_options, return_int=False):
...
+
+
+def get_seed_list(packing_config_data, recipe_data):
+    # Returns a list of seeds to use for packing
+    if packing_config_data["randomness_seed"] is not None:
+        seed_list = packing_config_data["randomness_seed"]
+    elif recipe_data.get("randomness_seed") is not None:
+        seed_list = recipe_data["randomness_seed"]
+    else:
+        seed_list = None
```
rugeli, line 269: Nit: should we implement an early return here for `seed_list` when it's `None`? So that we don't need to check it again in L274?

rugeli (approved):
LGTM! I was able to run the test variable count recipe using different config file settings without encountering any issues.

meganrm (approved):

PR #216, opened 2023-12-20
Save gradient data as image

### LINKED ISSUE #199: Export grid distances and weights as a tiff file
## Use Case
We want to visualize distances and weights stored in the grid as a tiff file to explore 3D slices. Simularium example:
image

## Solution

1. Create array of values stored in the grid. This array will have the same size as the grid (may be high/low resolution depending on grid spacing)
2. Use the `ImageWriter` class to write these values to an ome tiff image.

mogres:
Note: Merge #215 before this PR

Problem
=======
We want to export grid distances and weights as 3D images.
Closes #199 

Solution
========
Used the `ImageWriter` class to export gradient distances and weights as 3D tiff images.

## Type of change

Change summary:
---------------
* add option to save grid data as image 
* move voxelization code to environment
* add voxelization function for gradient distances and weights
* refactor image writer export code
* add recipe and config for testing

Steps to Verify:
----------------
1. `pack -r cellpack/tests/recipes/v2/test_nested_mesh_gradient.json -c cellpack/tests/packing-configs/test_gradient_data_image.json`
2. Open the output folder and navigate to `figures/gradient_data_figures` to view the exported image.

Screenshots:
-----------------------
Distances (mean intensity projection):
image

Weights (mean intensity projection):
image

-----------------------
1. `ImageWriter.py`
2. `Gradients.py`


mogres:
Use `nbvv` or volume viewer to visualize the outputs of this data.

cellpack/tests/packing-configs/test_gradient_data_image.json
```diff
@@ -0,0 +1,24 @@
...
+    "use_periodicity": false,
+    "show_sphere_trees": false,
+    "load_from_grid_file": true,
+    "save_converted_recipe": true,
+    "image_export_options": {
+        "hollow": false,
+        "voxel_size": [1,1,1],
+        "projection_axis": "z"
+    },
+    "save_gradient_data_as_image": true
```
meganrm, line 23: if this is a new config setting it should also have a default setting in the config loader. 
I also think, for clarity, any settings here that are defaults should be removed. The configs are ballooning a bit

cellpack/bin/pack.py
```diff
@@ -70,6 +70,9 @@ def pack(recipe, config_path=None, analysis_config_path=None):
             recipe_version=recipe_data["version"],
             image_export_options=packing_config_data.get("image_export_options"),
             parallel=packing_config_data.get("parallel", False),
+            save_gradient_data_as_image=packing_config_data.get(
+                "save_gradient_data_as_image", False
+            ),
```
meganrm, line 75: this is starting to make me think we should just sent the config into analysis instead of unpacking values here

meganrm (approved):
looks good to me, just one request that you add the new config setting to the config loader so we don't have mystery settings

rugeli (approved):
I ran the test recipe and was able to view the image without any problems. And I'm on board with adding `save_gradient_data_as_image` into config loader. Great work!

cellpack/tests/packing-configs/test_gradient_data_image.json
mogres: Removed the default entries from the config file and added the new setting to `config_loader.py`

cellpack/bin/pack.py
mogres: Created an issue #227 to track this

PR #223, opened 2024-01-19
Fix: prevent duplicate uploads of single packings

rugeli:
Problem
=======
When `"save_analyze_result"` is set to `True` in the config file, a single packing recipe will be uploaded twice

In `Analysis.py` at L2490, we made an update to the `doloop` function to call `save_as_simularium` only when `number_of_packings` > 1

## Type of change

Steps to Verify:
----------------
1. `pack -r examples/recipes/v2/one_sphere.json -c examples/packing-configs/debug.json `


mogres (approved):
LGTM!

meganrm (approved):

PR #224, opened 2024-01-26
change extended to true

meganrm:
Problem
=======
trimesh grid was not putting surface ingredients in the right spot with spheres as containers. 

Solution
========
raytrace worked and the only difference was "ex=True" so changed that. this should probably be dug into further in the future

## Type of change

Change summary:
---------------
* Tidy, well formulated commit message
* Another great commit message
* Something else I/we did

Steps to Verify:
----------------
1. A setup step / beginning state
1. What to do next
1. Any other instructions
1. Expected behavior
1. Suggestions for testing

Screenshots (optional):
-----------------------

-----------------------
1. main file/entry point
2. other important file

Thanks for contributing!

### DIFF
--- a/cellpack/autopack/Compartment.py
+++ b/cellpack/autopack/Compartment.py
@@ -1367,7 +1367,7 @@ class Compartment(CompartmentList):
 
         off_grid_surface_points = surface_points_in_bounding_box
 
-        ex = False  # True if nbGridPoints == len(idarray) else False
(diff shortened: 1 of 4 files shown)


mogres (approved):
Looks good to me! We should probably create an issue to track what `ex=True` is doing under the hood.

rugeli (approved):

PR #226, opened 2024-01-29
remove pandas3d

### LINKED ISSUE #219: Consider removing panda3d dependency?
## Use Case
Our current default packing method `spheresSST` does not require `panda3d` which adds extra overhead and error messages during import. 

Running a test recipe after disabling `panda3d` (by commenting out lines of code where it was used) resulted in a speedup of ~30%. This can be significant when generating a large number of packings.

## Solution
Remove functions and imports that rely on `panda3d`. Currently, there is some code that uses it for real time display which is no longer an available feature.

## Alternatives
Only remove `panda3d` from the `spheresSST` workflow while still keeping it in other places if needed.

meganrm:
Problem
=======
pandas3d slow to install and we're not currently using it. Closes #219 

Solution
========
removed it 

## Type of change

Steps to Verify:
----------------
1. all existing recipes should work as expected


rugeli:
Looks good to me! Here are my observations: 
1. Local v2 recipes tested with either `jitter` or `spheresSST` as the `place_method` were successfully packed:

- gradients.json
- nested.json
- one_sphere.json
- partner_packing_fv2.1.json
- peroxisomes.json
- sphere_tree.json

2. V1 recipes were working as expected as well.
3. We should delete the unused recipe `examples/recipes/v2/peroxisomes_gradient.json`. Although I believe Saurabh had previously removed it, it appears to still be present in `main`
4. A thought unrelated to this PR: are we using or planning to use `place_method == spring`?

docs/RECIPE_SCHEMA.md
```diff
@@ -94,16 +94,14 @@ Display packing in realtime (slow)
 
 ### place_method
 
-_Optional `enum`_. One of `"jitter"`, `"spheresSST"`, `"pandaBullet"`.. Default: `"jitter"`.
+_Optional `enum`_. One of `"jitter"`, `"spheresSST"`.. Default: `"jitter"`.
```
mogres, line 97: Is the current default `spheresSST`?

cellpack/autopack/Environment.py
```diff
@@ -2701,324 +2648,26 @@ def setupOctree(
                 self.grid.getRadius(), helper=helper
             )  # Octree((0,0,0),self.grid.getRadius())   #0,0,0 or center of grid?
```
mogres, line 2650: This function used to take so long to run! Good to have it removed

mogres (approved):
Great update! Nice to get rid of unused functionality.

rugeli (approved):

mogres:
We can merge this one if ready, @meganrm

PR #230, opened 2024-02-13
enhanced simularium url and firebase recipe path validation

### LINKED ISSUE #212: validate firebase recipe packing paths with explicit error messaging
## Use Case
the current command structure for packing firebase recipes, `pack -r firebase:recipes/[FIREBASE-RECIPE-ID] -c [PATH-TO-CONFIG]`, requires manual input that includes typing and pasting, which is prone to errors

## Solution
- minimize the need for manual typing and pasting in the command line
- provide a clear error message that pinpoints any typos in the recipe path

## Alternatives

rugeli:
Problem
=======
1. query parameters such as `AWSAccessKeyId` were not being stripped from the url when opening in a new tab 
2. closes #212

Solution
========
- removed query params from the url not only in firebase but also when opening a new tab
- a new field has been added to `recipes/` in firebase, making it easier to copy the firebase recipe path for packing
Screenshot 2024-02-12 at 4 31 27 PM

- enhanced validation for the input path to check each part of the path and provided with corresponding error messages  

## Type of change

Steps to Verify:
----------------
We have tested several cases:
- fetching a recipe from an incorrect database (e.g. uploading recipe A to `dev` but fetching it from `staging`)
- misnaming any part of the path including database name, collection name or recipe id  

some of the expected behaviors: 
- run `pack -r wrong_db_name:recipes/one_sphere_v_1.0.0`, get error:
Screenshot 2024-02-12 at 4 20 12 PM

- run `pack -r firebase:wrong_collection_name/one_sphere_v_1.0.0`, get error:
Screenshot 2024-02-12 at 4 21 42 PM

- run `pack -r firebase:recipes/wrong_id` or choosing the wrong database, get error:
Screenshot 2024-02-12 at 4 22 31 PM


cellpack/autopack/AWSHandler.py
```diff
@@ -89,10 +89,11 @@ def create_presigned_url(self, object_name, expiration=3600):
...
 
-    def save_file(self, file_path):
+    def save_file_and_get_url(self, file_path):
         """
         Uploads a file to S3 and returns the presigned url
         """
         file_name = self.upload_file(file_path)
         if file_name:
-            return file_name, self.create_presigned_url(file_name)
+            base_url = self.create_presigned_url(file_name).split("?")[0]
```
mogres, line 98: Nice!

mogres (approved):
Looks good!

cellpack/autopack/AWSHandler.py
meganrm: I might push this into the the `create_presigned_url` function so the function returns a url that the next function knows it can use without any adjustments. I also think this is a good bug for some tests. Ie make a list of urls that you know would be a problem and make sure we catch all of them.

rugeli:
Recent changes since the last review:
- added function `is_url_valid` within `AWSHandler` class
- added a new test file: `cellpack/tests/test_aws_handler.py`

meganrm (approved):

mogres (approved):
Thanks for adding tests!

PR #231, opened 2024-02-14
fix: convert "type" Enums to strings in v1 recipe for firebase upload duplication check

rugeli:
Problem
=======
Found a bug: a v1 recipe could be uploaded to firebase more than once 

This issue arises because, during the version migration process, the "type" attribute of an object is converted into an Enum, whereas it is stored as a string in firebase database. So our duplication checking logic always find a difference in data types.

difference:
```
{
  "type_changes": {
    "root['type']": {
      "old_type": "<class 'str'>",
      "new_type": "<enum 'INGREDIENT_TYPE'>",
      "old_value": "single_sphere",
      "new_value": "<INGREDIENT_TYPE.SINGLE_SPHERE: 'single_sphere'>"
    }
  }
}

```

Solution
========
I added a value check for enum in the existing method `prep_data_for_db` within `DBRecipeHandler` 

## Type of change

Steps to Verify:
----------------
1. in `DBRecipeHandler.py`, comment out the `return` in `DBUploader > upload_recipe()` to let deeper system checks into objects and compositions
2. run `upload -r examples/recipes/v1/one_sphere.json` to upload the recipe to your dev database
3. repeat the upload command to verify that the duplication check correctly identifies each object and composition, preventing duplicate uploads

### DIFF
--- a/cellpack/autopack/DBRecipeHandler.py
+++ b/cellpack/autopack/DBRecipeHandler.py
@@ -1,4 +1,5 @@
 import copy
+from enum import Enum
 
 from deepdiff import DeepDiff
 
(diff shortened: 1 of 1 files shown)


mogres (approved):
Good catch! Apologies it took so long to review.

meganrm (approved):

Issue #227, opened 2024-02-03
Send in the config data into Analysis.py


mogres:
## Use Case
From [this](https://github.com/mesoscope/cellpack/pull/216#discussion_r1467115271) comment

We are sending many arguments into `Analysis.doloop` that are unpacked from config data. Consider sending in the entire config_data dictionary and unpacking inside the function

## Solution
Send in `config_data` as a single argument into doloop. Unpack individual arguments inside the function

PR #232, opened 2024-02-21
Feature/send config data to analysis

mogres:
Problem
=======
solves #227 

Solution
========
Send in `recipe_data` and `config_data` into the `doloop` function


meganrm (approved):

rugeli (approved):
LGTM!

PR #240, opened 2024-03-20
moved mdutils to analysis requirements

meganrm:
Problem
=======
the install failed because it needed mdutils 

Solution
========
mdutils is only used in the analysis code so I moved the requirement there 

## Type of change


mogres (approved):
Looks like `mdutils` was the only missing dependency. Hopefully once we re-release to pip we can run an example packing

PR #241, opened 2024-03-26
Automated cleanup of expired result metadata in staging firebase database 

### LINKED ISSUE #195: delete expired S3 metadata in firebase
## Use Case
To avoid unnecessary costs, we want to delete result files from our S3 bucket that older than 6 months 

## Solution
set up a github action that runs periodically to identify and remove old files

## Alternatives

rugeli:
Problem
=======
closes #195 

Solution
========
- scheduled a github action to weekly cleanup results metadata in the staging firebase database 
- added github secrets for the staging db credentials to our repo, named `FIREBASE_TOKEN` and `FIREBASE_EMAIL`  
- implemented the logic in `FirebaseHandler` and `DBRecipeHandler` to verify the existence of packing results in the S3 bucket, if not found, remove the corresponding metadata from firebase

## Type of change

Steps to Verify:
----------------
1. View the most recent cleanup github action run for testing at this [link](https://github.com/mesoscope/cellpack/actions/runs/8440729655/job/23118306690)

Or test the main functions locally:
1. pack a test recipe, and run `python cellpack/bin/cleanup_tasks.py`. Temporarily adjust the condition from `if result_age.days > 180 and not self.validate_existence(result_data["url"]):` to something like `if result_age.total_seconds() < 1000`. To check if the metadata of your test recipe gets deleted as expected


.github/workflows/cleanup-firebase.yml
```diff
@@ -0,0 +1,29 @@
+name: Cleanup Firebase Metadata
+
+on:
+    schedule:
```
rugeli, line 4: Currently, this job is scheduled to run weekly. We can add additional triggers or modify the frequency of the job if needed.

.github/workflows/cleanup-firebase.yml
```diff
@@ -0,0 +1,29 @@
...
+        - cron: "24 18 * * 1" # Runs at 18:24 UTC every Monday
```
ascibisz, line 5: out of curiosity, is there a reason you picked this time? it feels very specific

rugeli: I simply copied the line from another workflow `ci.yml`, this comment may answer the question:
```
 # (Since these CRONs are used by a lot of people -
 # let's be nice to the servers and schedule it _not_ on the hour)
```

cellpack/autopack/DBRecipeHandler.py
```diff
@@ -376,6 +378,37 @@ def should_write(self, db, grad_name):
...
+        else:
+            print("No results found in the database.")
+
+    def validate_existence(self, url):
+        """
+        Validate the existence of an S3 object by checking if the URL is accessible.
+        Returns True if the URL is accessible.
+        """
+        response = requests.head(url)
+        return response.status_code == 200
```
ascibisz, line 409: [small nit]: perhaps replace the `200` with `requests.codes.ok`, just in case someone is unfamiliar with http response codes?

.github/workflows/cleanup-firebase.yml
ascibisz: ahh gotcha!

Follow up question: I'm still not super familiar with firebase, will the performance of reading to / writing from the database be interrupted while this job is running? If so, it could be nice to have the job run in the middle of the night or sometime when we wouldn't be using it (18:24 UTC is like 11:24am pacific time)

cellpack/autopack/DBRecipeHandler.py
```diff
@@ -376,6 +378,37 @@ def should_write(self, db, grad_name):
...
+            for result in results:
+                result_data = self.db.doc_to_dict(result)
+                result_age = current_utc - result_data["timestamp"]
+                if result_age.days > 180 and not self.validate_existence(
+                    result_data["url"]
+                ):
+                    self.db.delete_doc("results", self.db.doc_id(result))
+            print("Results cleanup complete.")
+        else:
+            print("No results found in the database.")
```
ascibisz, line 401: Do you have a preference between `print` and `logging` for this repo? I saw in `FirebaseHandler`, you're changing some `print`s to `logging.info`s, so I was just curious

.github/workflows/cleanup-firebase.yml
rugeli: Since this job is to check data under the `results` collection, it shouldn't be interrupting any firebase's performances. But yeah we can totally schedule a time during non-working hours

cellpack/autopack/DBRecipeHandler.py
rugeli: Good question! I'd say we should use `logging` over `print` for most scenarios, but for statements like these, I lean towards a cleaner message like `Results cleanup complete.` rather than the logging `2024-03-26 10:57:37,029 - root - INFO - FirebaseHandler.py:205 -           delete_doc(): successfully deleted path`.

I'm open to suggestions and happy to hear better approaches on this!

.github/workflows/cleanup-firebase.yml
ascibisz: Seems like it shouldn't be an issue then, thanks for clarifying!

cellpack/autopack/DBRecipeHandler.py
ascibisz: Gotcha, that makes sense to me!

ascibisz (approved):
I'm still learning my way around the cellpack repo, but this looks good to me!

.github/workflows/cleanup-firebase.yml
rugeli: only `schedule` should be suffice for now

meganrm (approved):
looks great, I do agree with comment Alli made that I thumbs up-ed

PR #247, opened 2024-05-08
Fix github action job: Version 3.9 with arch x64 not found

rugeli:
Merge this pr first

Problem
=======
CI/CD issue: `Error: Version 3.9 with arch x64 not found`

Solution
========
~~Made a few adjustments, and turns out using python 3.10 - currently the most stable version for cellPACK.~~
updated `checkout` and `setup-python`

## Type of change

* Bug fix and version update (non-breaking change which fixes an issue)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.10]
+        python-version: ["3.10"]
         os: [ubuntu-latest, windows-latest, macOS-latest]
 
     steps:
@@ -53,7 +53,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v5
       with:
-        python-version: 3.10
+        python-version: "3.10"
     - name: Install Dependencies
       run: |
         python -m pip install --upgrade pip
@@ -80,7 +80,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v1
       with:
-        python-version: 3.10
+        python-version: "3.10"
     - name: Install Dependencies
       run: |
         python -m pip install --upgrade pip
@@ -89,5 +89,5 @@ jobs:
       run: |
         python setup.py sdist bdist_wheel
     - name: Publish to PyPI
-      uses: pypa/gh-action-pypi-publish@release/v1 
+      uses: pypa/gh-action-pypi-publish@release/v1
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,13 +26,13 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [3.9]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
         os: [ubuntu-latest, windows-latest, macOS-latest]
 
     steps:
-    - uses: actions/checkout@v1
+    - uses: actions/checkout@v4
     - name: Set up Python ${{ matrix.python-version }}
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v5
       with:
         python-version: ${{ matrix.python-version }}
     - name: Install Dependencies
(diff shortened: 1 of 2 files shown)


mogres:
Is there any reason we are using Python 3.10 for testing? The base cellpack environment still uses Python 3.9

rugeli:
> Is there any reason we are using Python 3.10 for testing? The base cellpack environment still uses Python 3.9

Because one of our github actions failed, with the error described in the title, it seems that the version of `ubuntu-latest` is the cause. 

To solve the issue, we can either specify an older ubuntu version(20.04) or upgrade the python version, and I chose the latter solution in this pr. I tried a few python versions but ran into issues with deprecations and `module not found`, version 3.10 finally passed all the tests. 

We may need to upgrade the base environment to python 3.10 soon to maintain consistency.

.github/workflows/ci.yml
```diff
@@ -53,7 +53,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v5
       with:
-        python-version: 3.9
+        python-version: "3.10"
```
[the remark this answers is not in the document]
rugeli: Oh I just realized sharing values across jobs might require some tweaks, like adding another job to define and output the python version and using `needs` in other jobs. The way we have it now looks cleaner.

mogres:
> To solve the issue, we can either specify an older ubuntu version(20.04) or upgrade the python version, and I chose the latter solution in this pr. I tried a few python versions but ran into issues with deprecations and `module not found`, version 3.10 finally passed all the tests.
> 
> We may need to upgrade the base environment to python 3.10 soon to maintain consistency.

While I do agree we should move to python 3.10 eventually, I am not sure if this is the right approach. The ideal path would be to first create a branch to migrate cellPACK to Python 3.10, check whether this breaks anything beyond our unit tests and then upgrade our CI workflows, requirements, etc. to use the new python version.

Another reasoning for sticking with python 3.9 is because the cellPACK package is a dependency in other analysis packages including `simulariumio` and `cellpack-analysis` and we want to make sure these work as expected when we eventually release a newer version with python 3.10.

For the time being, I would recommend using Ubuntu 20.04 in our tests to be compatible with the base cellPACK version. If that is not feasible, we can discuss ways to ensure that the tests are run on the same python version as the base package in one of the WG meetings.

.github/workflows/ci.yml
```diff
@@ -80,7 +80,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v1
       with:
-        python-version: 3.9
+        python-version: "3.10"
```
mogres, line 83: Updating the setup-python version here might also be useful

rugeli:
> > To solve the issue, we can either specify an older ubuntu version(20.04) or upgrade the python version, and I chose the latter solution in this pr. I tried a few python versions but ran into issues with deprecations and `module not found`, version 3.10 finally passed all the tests.
> > We may need to upgrade the base environment to python 3.10 soon to maintain consistency.
> 
> While I do agree we should move to python 3.10 eventually, I am not sure if this is the right approach. The ideal path would be to first create a branch to migrate cellPACK to Python 3.10, check whether this breaks anything beyond our unit tests and then upgrade our CI workflows, requirements, etc. to use the new python version.
> 
> Another reasoning for sticking with python 3.9 is because the cellPACK package is a dependency in other analysis packages including `simulariumio` and `cellpack-analysis` and we want to make sure these work as expected when we eventually release a newer version with python 3.10.
> 
> For the time being, I would recommend using Ubuntu 20.04 in our tests to be compatible with the base cellPACK version. If that is not feasible, we can discuss ways to ensure that the tests are run on the same python version as the base package in one of the WG meetings.

I see your concern, sure I'll try a different approach and we can discuss it at our next meeting.

.github/workflows/ci.yml
```diff
@@ -80,7 +80,7 @@ jobs:
     - name: Set up Python
       uses: actions/setup-python@v1
```
mogres, line 81:
```suggestion
      uses: actions/setup-python@v5
```

mogres (commented):
Maybe updating the setup-python version would help?

.github/workflows/ci.yml
rugeli: you are absolutely right!

mogres: Yay!

mogres (approved):

meganrm (approved):

PR #248, opened 2024-05-08
call path validation when initiating the packing for remote recipes

### LINKED ISSUE #245: Cannot pack firebase recipes due to missing "composition" key
## Description
On the main branch, packing a firebase recipe results in an error. Here are the steps I followed:

### Local recipe works as expected
image

### Recipe is available on firebase
image

### Packing the recipe from firebase results in an error
image

## Expected Behavior
The firebase recipe should run similar to the local recipe

rugeli:
Problem
=======
closes #245 

Solution
========
Added the function call back for remote packings.
Expected behaviors in merged pr #230  

## Type of change

### DIFF
--- a/cellpack/autopack/__init__.py
+++ b/cellpack/autopack/__init__.py
@@ -388,6 +388,7 @@ def load_file(filename, destination="", cache="geometries", force=None):
             db = DATABASE_IDS.handlers().get(database_name)
             initialize_db = db()
             db_handler = DBRecipeLoader(initialize_db)
+            db_handler.validate_input_recipe_path(filename)
             recipe_id = file_path.split("/")[-1]
             db_doc, _ = db_handler.collect_docs_by_id(
                 collection="recipes", id=recipe_id


rugeli:
Quickest way to get the remote recipe path is to copy and paste the field `recipe_path` within `recipes` collection.
Screenshot 2024-05-08 at 1 30 35 PM

mogres (approved):
Thanks for the quick update!

I was able to run the `one_sphere` recipe after deleting and reuploading it. The previous version did not have the `recipe_path` field and I wasn't able to update it by reuploading since a recipe with the same name was already in the firestore. Maybe some validation to update the recipe if the format/saved fields have changed might be useful?

rugeli:
> Thanks for the quick update!
> 
> I was able to run the `one_sphere` recipe after deleting and reuploading it. The previous version did not have the `recipe_path` field and I wasn't able to update it by reuploading since a recipe with the same name was already in the firestore. Maybe some validation to update the recipe if the format/saved fields have changed might be useful?

The Firebase system is still in active development, and as we're refining it, we may need to delete old test recipes periodically until things stabilize. I agree that adding validation in the future to handle updates, especially when fields or formats are missing, will be useful! Thanks for testing the database!

meganrm (approved):

PR #250, opened 2024-05-09
Cache pip dependencies in github actions

### LINKED ISSUE #249: Cache python installation in github actions
## Use Case
Python installation takes a long time while running our tests through github actions. We can cache the python version and update as needed to speed up the workflows.

Refer [here](https://github.com/aics-int/nuc-morph-analysis/pull/378) for an example.

More information on caching [here](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows)

## Solution
Add `cache: true` in the install dependencies step

mogres:
Problem
=======
Closes #249 

Solution
========
Added `cache: "pip"` to github action workflows that install cellpack dependencies

## Type of change

Steps to Verify:
----------------
Checking runtime for github actions, we get a speedup of a few minutes on MacOS and Windows tests

Screenshots:
-----------------------
### Before (from PR #247)
image

### After:
image


rugeli (approved):
We might want to consider upgrading `setup.py` to `pyproject.toml` sometime down the line. Otherwise, lgtm!

mogres:
> We might want to consider upgrading `setup.py` to `pyproject.toml` sometime down the line. Otherwise, lgtm!

Agreed!

I think a good way to do this is by using [PDM](https://pdm-project.org/en/stable/) to manage dependencies and create a synced `pyproject.toml` that is stable across operating systems.

meganrm (approved):

PR #252, opened 2024-05-24
Remove pyembree

### LINKED ISSUE #251: Remove pyembree dependency
## Use Case
The Pyembree package is not maintained and also not used in our current mesh packings. It also forces pinning of other dependencies (numpy, cython, trimesh, etc) to lower versions. 

## Solution
Remove pyembree from cellPACK requirements

## Testing
* Remove pyembree
* Install cellpack requirements and run some test packings with meshes

mogres:
Problem
=======
Removes unused/deprecated pyembree package
Closes #251 

Solution
========
Updated requirements to remove functions using pyembree

Steps to Verify:
----------------
Pack the peroxisome surface gradient example recipe with old and new requirements to compare grid creation and packing time. 

### with pyembree
image

### without pyembree
image

We see 8-10% improvement in grid creation and packing speed.


meganrm (approved):

rugeli (approved):
lgtm, thanks for the update!

PR #255, opened 2024-05-29
Feature/add plots to result

### LINKED ISSUE #254: Adding plots to result file
## Use Case
we'd like to embed plot visualizations for packings to facilitate a more comprehensive analysis from the simularium result files.

rugeli:
Problem
=======
closes #254 

Solution
========

- created a class to handle plot data(scatter plots and histograms) and added all plots to one list
- in the function `histogram` within `Analysis.py`, added the logic to insert plot data into result file. Set the argument `add_to_result` to true and set `save_png` to false, by default 
- in `simularium_helper`, added plot data to the trajectory converter and had simularium io process the plot display on the website     

with @meganrm @mogres @ascibisz and Joe

## Type of change

Steps to Verify:
----------------
1. Pack any recipes with the `save_analyze_result` config set to `true` to see if the plots display next to the result as expected 

Screenshots (optional):
-----------------------
Screenshot 2024-05-29 at 1 22 15 PM

### DIFF
--- a/cellpack/autopack/Environment.py
+++ b/cellpack/autopack/Environment.py
@@ -2178,9 +2178,7 @@ class Environment(CompartmentList):
             pbar.close()
         self.log.info("time to fill %d", t2 - t1)
         all_objects = self.prep_molecules_for_save(distances, free_points, nbFreePoints)
-        print("SELF.result", self.saveResult)
         if self.saveResult:
-            print("SAVING FROM ENV")
             self.save_result(
                 free_points,
                 distances=distances,
--- a/cellpack/autopack/Environment.py
+++ b/cellpack/autopack/Environment.py
@@ -143,7 +143,11 @@ class Environment(CompartmentList):
         self.name = name
         self.version = recipe.get("version", "default")
         # saving/pickle option
-        self.saveResult = "out" in config
+        self.saveResult = (
+            "out" in config
+            and not config["save_analyze_result"]
+            and not config["number_of_packings"] > 1
+        )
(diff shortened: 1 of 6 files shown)


cellpack/autopack/Environment.py
```diff
@@ -143,7 +143,11 @@ def __init__(self, config=None, recipe=None):
         self.name = name
         self.version = recipe.get("version", "default")
         # saving/pickle option
-        self.saveResult = "out" in config
+        self.saveResult = (
+            "out" in config
+            and not config["save_analyze_result"]
```
mogres, line 148: Why do we have these additional conditions here? Wouldn't this prevent saving simularium files completely if there is more than one packing?

mogres (approved):
Really cool feature! The `PlotData` class is super useful :) I was able to run the `test_spheres` recipe and generate simularium files with plots on the side!

Your PR description asks to set `show_grid_plot` config to `true` but I don't see it checked anywhere. Do you mean `save_analyze_result`?

cellpack/autopack/Environment.py
meganrm: it's because the saving happens from the Analysis class instead of the Env class

meganrm (approved):

rugeli:
> Really cool feature! The `PlotData` class is super useful :) I was able to run the `test_spheres` recipe and generate simularium files with plots on the side!
> 
> Your PR description asks to set `show_grid_plot` config to `true` but I don't see it checked anywhere. Do you mean `save_analyze_result`?

ah right, I meant `save_analyze_result`, corrected!

PR #256, opened 2024-05-29
Load grid from URL

### LINKED ISSUE #253: Load grid files from URL
## Purpose
We want to be able to load grid files from remote databases such as s3 or google drive. This will enable us to run the cellPACK analysis pipeline fully on the cloud without having to generate large grid files which take a few minutes per grid.

## Approach
We currently have ways to download mesh files from a URL and store them in a local cache. We can reuse these functions for grid files.

## Considerations
Grid files can be large (up to a few GBs). We should consider adding a `clean` binary command to clear the local cellPACK grid cache.

## Testing
Pack peroxisomes in the average cell and nuclear shape by loading a grid file from s3.

mogres:
Problem
=======
Closes #253 

Solution
========
* Updated grid file loading to use URLs
* Moved cache to local repo clone
* Added functionality to clean cache

## Type of change

Steps to Verify:
----------------
### Run Packing
```python
pack -r cellpack/tests/recipes/v2/test_url_load.json -c cellpack/tests/packing-configs/test_url_load_config.json
```
### Details
The packing should download and save a file `test_s3_mesh_test_s3_mesh_config_1.0.0_grid.dat` to `.cache/grids/` folder at the top level of the repository
### Clean Cache
Since some grid files can be large, we want to clean the cache periodically

#### Option 1
Set the `clean_grid_cache` option to `true` in the config file. This will delete the downloaded grid file after the packing is completed.

#### Option 2 (Destructive)
Run `python cellpack/bin/clean.py`. 
**WARNING**: This will remove all the files in the cache!

### DIFF
--- a/cellpack/autopack/__init__.py
+++ b/cellpack/autopack/__init__.py
@@ -36,11 +36,9 @@ AF
 """
 import logging
 import logging.config
-import sys
 import os
 import re
 import shutil
-from os import path, environ
 import getpass
 from pathlib import Path
 import urllib.request as urllib
@@ -61,7 +59,9 @@ afdir = Path(os.path.abspath(__path__[0]))
 os.environ["NUMEXPR_MAX_THREADS"] = "32"
 
 ###############################################################################
-log_file_path = path.join(path.dirname(path.abspath(__file__)), "../logging.conf")
+log_file_path = os.path.join(
+    os.path.dirname(os.path.abspath(__file__)), "../logging.conf"
+)
 logging.config.fileConfig(log_file_path, disable_existing_loggers=False)
 log = logging.getLogger("autopack")
 log.propagate = False
--- a/cellpack/autopack/__init__.py
+++ b/cellpack/autopack/__init__.py
@@ -290,9 +290,19 @@ def parse_s3_uri(s3_uri):
 def download_file(url, local_file_path, reporthook):
     if is_s3_url(url):
         # download from s3
+        # bucket_name, folder, key = parse_s3_uri(url)
+        # s3_handler = DATABASE_IDS.handlers().get(DATABASE_IDS.AWS)
+        # s3_handler = s3_handler(bucket_name, folder)
+        s3_client = boto3.client("s3")
         bucket_name, folder, key = parse_s3_uri(url)
-        s3_handler = DATABASE_IDS.handlers().get(DATABASE_IDS.AWS)
-        s3_handler = s3_handler(bucket_name, folder)
+        try:
+            s3_client.download_file(bucket_name, f"{folder}/{key}", local_file_path)
+            print("File downloaded successfully.")
+        except botocore.exceptions.ClientError as e:
+            if e.response["Error"]["Code"] == "404":
+                print("The object does not exist.")
+            else:
+                print("An error occurred while downloading the file.")
 
     elif url_exists(url):
         try:
--- a/cellpack/autopack/__init__.py
+++ b/cellpack/autopack/__init__.py
@@ -77,24 +77,9 @@ def make_directory_if_needed(directory):
 # #Setup autopack data directory.
 # ==============================================================================
 # the dir will have all the recipe + cache.
-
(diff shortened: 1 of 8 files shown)


cellpack/autopack/__init__.py
```diff
@@ -109,23 +96,24 @@ def url_exists(url):
...
 # ==============================================================================
 
-
-cache_results = appdata / "cache_results"
-cache_geoms = appdata / "cache_geometries"
-cache_sphere = appdata / "cache_collisionTrees"
-cache_recipes = appdata / "cache_recipes"
+cache_results = appdata / "results"
+cache_geoms = appdata / "geometries"
+cache_sphere = appdata / "collisionTrees"
```
meganrm, line 101: curious: why this change?

cellpack/autopack/__init__.py
```diff
@@ -271,8 +300,19 @@ def download_file(url, local_file_path, reporthook):
         raise Exception(f"Url does not exist {url}")
 
 
+# def is_full_url(file_path):
+#     return file_path.find("http") != -1 or file_path.find("ftp") != -1
+
```
meganrm, line 305: delete?

meganrm (approved):
looks good, just some non blocking comments

cellpack/autopack/__init__.py
```diff
@@ -261,8 +250,49 @@ def updateReplacePath(newPaths):
         REPLACE_PATH[w[0]] = w[1]
 
 
+def download_file_from_s3(s3_uri, local_file_path):
```
rugeli, line 253: it seems like this function does the same thing as `def download_file()` and isn't being used. we can keep one of them and remove the other

cellpack/autopack/__init__.py
```diff
@@ -261,8 +250,49 @@ def updateReplacePath(newPaths):
...
+    folder = "/".join(parts[1:-1])
+    key = parts[-1]
+
+    return bucket_name, folder, key
+
+
 def download_file(url, local_file_path, reporthook):
-    if url_exists(url):
+    if is_s3_url(url):
+        # download from s3
```
rugeli, line 280: These steps you've commented out are indeed the correct way to initiate the s3 client, we have functionality in `AWSHandler` that handles the initiation of clients and manages multiple existing clients. I'd suggest moving either `download_file` or `download_file_from_s3` to `AWSHandler` to keep aws related util functions more organized and avoid potential client conflicts in the future. What do you think?

rugeli (commented):
The cache cleaning and grid file loading parts look great! For the S3 part, I created a branch to dig around and moved `download_file_from_s3` to AWSHandler, and it works the same as this branch does. Here is the comparison [link](https://github.com/mesoscope/cellpack/compare/feature/load_mesh_from_s3...refactor/load_from_s3?expand=1) for you to take a look at. We can also chat about tomorrow if you like:)

cellpack/autopack/__init__.py
mogres: Good point!

cellpack/autopack/__init__.py
mogres: Makes sense! It looks like you have moved these files in your branch already? In that case, I will leave these S3 related functions here so the refactoring in your branch doesn't have merge conflicts.

mogres (commented):
Thanks for taking a look @rugeli! 
I suggest we merge this branch with the s3 related functions in `__init__.py` and then move them to `AWSHandler` in a later PR from the branch you linked to earlier. Would that work? The refactoring from the other branch looks good to me!

> Nit: the file name has some repeated parts in it: saved to out/test/test_s3_mesh/spheresSST/results_test_s3_mesh_test_s3_mesh_config_1.0.0_seed_0.simularium. We might want to reconstruct it.

This is because my recipe is called `test_s3_mesh` and the config is called `test_s3_mesh_config`. I'll rename these for clarity.

rugeli (approved):

PR #268, opened 2024-06-24
Adding `grid_file_path` to recipe metadata for uploads 

### LINKED ISSUE #266: Upload grid file path to firebase on upload
## Use Case
We want to add a field `grid_file_path` to store grid file link in the top-level info of a recipe in firebase

rugeli:
Problem
=======
closes #266 

Solution
========
- relocated function: removed `get_only_recipe_metadata` from `recipe_loader.py`, and added `get_recipe_metadata` in `upload.py` 
- modified the function to explicitly check for and extract the `grid_file_path` if it exists in recipe data

## Type of change

Steps to Verify:
----------------
1. Upload a recipe that includes the `grid_file_path` to firebase. Run `upload -r cellpack/tests/recipes/v2/test_url_load.json`
2. Verify that the recipe contains all the expected fields. 

Note: At this stage of development, we'll need to delete any existing recipe in firebase before re-uploading a recipe with the same name. Discussions are ongoing about enabling options to overwrite or update existing recipes

### DIFF
--- a/cellpack/bin/upload.py
+++ b/cellpack/bin/upload.py
@@ -10,7 +10,7 @@ from cellpack.autopack.loaders.recipe_loader import RecipeLoader
 
 def get_recipe_metadata(loader):
     """
-    Fetches and returns the metadata of the recipe for uploading
+    Extracts and returns essential metadata from a recipe for uploading
     """
     try:
         recipe_meta_data = {
--- a/cellpack/bin/upload.py
+++ b/cellpack/bin/upload.py
@@ -8,6 +8,25 @@ from cellpack.autopack.interface_objects.database_ids import DATABASE_IDS
 from cellpack.autopack.loaders.recipe_loader import RecipeLoader
 
 
+def get_recipe_metadata(loader):
+    """
+    Fetches and returns the metadata of the recipe for uploading
+    """
+    try:
+        recipe_meta_data = {
+            "format_version": loader.recipe_data["format_version"],
+            "version": loader.recipe_data["version"],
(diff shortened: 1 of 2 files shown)


cellpack/bin/upload.py
```diff
@@ -8,6 +8,25 @@
...
+    """
+    try:
+        recipe_meta_data = {
+            "format_version": loader.recipe_data["format_version"],
+            "version": loader.recipe_data["version"],
+            "name": loader.recipe_data["name"],
+            "bounding_box": loader.recipe_data["bounding_box"],
+            "composition": {},
+        }
+        if "grid_file_path" in loader.recipe_data:
```
mogres, line 23: Q: should we check for all keys in `loader.recipe_data` and add them to the `recipe_meta_data` (except `composition`)? This way we won't have to do a special check for `grid_file_path` or any new keys we add to the recipe down the line.

mogres (approved):
The changes look good! I was able to upload and run the example recipe from firebase.

cellpack/bin/upload.py
rugeli: Thank you for asking! I was thinking the same approach: iterating over recipe data and recording every key except for `objects`, `representations`, etc., to make the uploading more programmatic. However I ended up choosing to specify keys(for now) so we get to fully control which keys are displayed and needed in the metadata. Helping to prevent errors during data handling.

And you're right, if there are more new top-layer keys, we should think about a more dynamic way to manage it.

meganrm (approved):

PR #271, opened 2024-06-26
Specify weights while combining gradients

mogres:
Problem
=======
Allows multiple gradients to be added to a single ingredient in a tunable way.

Solution
========
Add the option to weight multiple gradients on a single ingredient

## Type of change

Change summary:
---------------
* Add the option `gradient_weights` to specify how to combine the multiple gradients

Steps to Verify:
----------------
Test weighted gradient combination:
```bash
pack -r cellpack/tests/recipes/v2/test_combined_gradient.json -c cellpack/tests/packing-configs/test_grid_plot_config.json
```

Screenshots:
-----------------------
 #### Weight map for weighted gradient combination
combined_gradient_weighted


cellpack/autopack/Gradient.py
```diff
@@ -176,13 +198,16 @@ def pick_point_for_ingredient(ingr, allIngrPts, all_gradients):
...
+                    for gradient_name in ingr.gradient:
+                        if gradient_name not in all_gradients:
+                            raise ValueError(
+                                f"Gradient {gradient_name} not found in gradient list"
+                            )
+                        gradient_list.append(all_gradients[gradient_name])
+
                     combined_weight = Gradient.get_combined_gradient_weight(
-                        gradient_list
+                        gradient_list, ingr.gradient_weights
```
rugeli, line 210: here might pass an empty `gradient_list` to `get_combined_gradient_weight` and result in an `IndexError`. Maybe add a check in `get_combined_gradient_weight`

rugeli (approved):
Except for one edge case handling, the code looks nice and clean to me!

cellpack/autopack/Gradient.py
mogres: In this case `gradient_list` should not be empty, since we check that `ingr.gradient` is a list with more than one element. All elements from `ingr.gradient` are added to `gradient_list` and an error is raised if there is a missing entry. I'll add a check to `get_combined_gradient_weight` just to make sure this is the case.

rugeli: oh, that makes sense! Thank you for the clarification and adding the check

cellpack/autopack/Gradient.py
```diff
@@ -91,6 +91,28 @@ def __init__(self, gradient_data):
...
+    @staticmethod
+    def update_ingredient_gradient(ingr, arguments):
+        """
+        Update the ingredient gradient
+        """
+        if (
+            "gradient" in arguments
+            and arguments["gradient"] != ""
+            and arguments["gradient"] != "None"
+        ):
```
meganrm, line 103: this seems like something that can be cleaned up in the recipe loader (not have multiple falsy possibilities)

mogres: Good idea! I added some validation and tests to check whether gradient information is entered correctly during ingredient creation from recipe arguments.

meganrm (approved):

Issue #283, opened 2024-08-22
Pass Recipe and Config into Docker Container


ascibisz:
## Use Case
* We need to be able to configure the Dockerfile to accept recipe and config parameters, and then pass those to the `pack` script as the `-r` and `-c` params
* Also, when running in docker, we don't want the question of staging or dev firebase to come up, should just default to staging

PR #285, opened 2024-08-22
Feature/pass recipe to docker

ascibisz:
Problem
=======
Cleaning up docker code so we are actually passing the recipe / config parameters specified in the commandline / as ENV variables into the docker container.
[Link to story or ticket](https://github.com/mesoscope/cellpack/issues/283)

Solution
========
- Updated docker-entrypoint.sh to actually read provided recipe and config file paths and pass them to the `pack` script
- Added .dockerignore file so we aren't copying unneeded files into the docker image
- Added an optional boolean `docker` for when we want to default to pointing to staging for firebase without prompting the user via commandline. To use, we can call the `pack` script from the commandline with the `-d` flag, i.e. `pack -r my_recipe.json -c my_config.json -d`

## Type of change

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,37 @@
+# Distribution / packaging
+build/
+develop-eggs/
+eggs/
+.eggs/
+*.egg-info/
+.installed.cfg
+*.egg
+__pycache__
+
+# Unit test / coverage reports
+htmlcov/
+.tox/
+.coverage
+.coverage.*
+.cache
+*.cover
+cellpack/tests/
+cellpack/bin/tests/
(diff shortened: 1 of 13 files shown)


rugeli (approved):
LGTM! We can update the documentation for triggering batch jobs once the gateway api is in place. If we want to rename variables like `use_docker`, we can handle that in a cleanup pr before merging into `main` from `staging`.

mogres (approved):
I was able to build the image but could not run it on the on-site dgx machines due to docker being unable to access my home directory. However, the implementation looks good to me!

PR #308, opened 2025-01-27
Fix/cleanup aws workflow

### LINKED ISSUE #304: Fix `awscli` installation in github actions

rugeli:
Problem
=======
closes #304 

Solution
========
- Replace the `Install AWS CLI` step with a script that installs CLI directly from AWS
- Included the `--update` flag to handle the existing installations

  
## Type of change

Steps to Verify:
----------------
1. This workflow should be triggered by closing a pull request 
2. I have tested this workflow on a testing branch, and AWS CLI was installed as expected

### DIFF
--- a/.github/workflows/cleanup-aws.yml
+++ b/.github/workflows/cleanup-aws.yml
@@ -1,9 +1,6 @@
 name: Cleanup AWS S3 Bucket
 
 on: 
-  push:
-    branches:
-      - fix/cleanup-aws-workflow
   pull_request:
     types: [closed]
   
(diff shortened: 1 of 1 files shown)


mogres:
Hi Ruge, the updates looks good! However, it looks like AWS CLI is preinstalled on github hoster runners as seen [here](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md) in the CLI tools section (more documentation [here](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners)). Can we check on the test branch whether removing the setup step has any effect? If so I would recommend just getting rid of the install step and using the preconfigured version of AWS CLI.

rugeli:
> Hi Ruge, the updates look good! However, it looks like AWS CLI is preinstalled on github hosted runners as seen [here](https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2404-Readme.md) in the CLI tools section (more documentation [here](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners)). Can we check on the test branch whether removing the setup step has any effect? If so I would recommend just getting rid of the install step and using the preconfigured version of AWS CLI.

Good point! We can simply rely on the preinstalled version. I've removed the installation step and added a step to check the version instead. It worked fine. Thank you for catching that!

mogres (approved):
Looks good! I like the idea of not having to download packages through `apt-get` as that can be sometimes unreliable

meganrm (approved):

PR #320, opened 2025-03-04
Refactor uploading system deduplication using hash 

### LINKED ISSUE #316: Optimize duplicate checking for recipe data upload in Firebase
## Use Case
When uploading recipe data to firebase, our system loops through all existing documents to check for duplicates. As the dataset grows, this approach becomes inefficient, leading to significant delays or even timeouts. 

## Solution

We need a more optimized duplicate-checking mechanism, potentially using a hash schema or indexed fields in firebase(TBD). 

 
This work can be broken into smaller tickets as we encounter various unexpected challenges. The goal is to ensure that our website users can save and run their modified recipes without performing issues.

rugeli:
Problem
=======
closes #316

Solution
========
- used `hashlib` to generate a unique hashes for each doc (any changes will produce a new hash)
- removed the `should_write` method from each doc class
-  updated unit tests
with @ascibisz 

- introduced new field `dedup_hash`
Screenshot 2025-03-04 at 1 32 29 PM

## Type of change
* Refactor (non-breaking change which adds functionality)

Steps to Verify:
----------------
1. Upload a recipe to your firebase dev db
2. Re-upload the same recipe to confirm the hash prevents duplicates
3. edit recipe's name and a nested data field, then upload again to test if new doc and new recipe are detected 
4. run `pytest`

### DIFF
--- a/cellpack/autopack/DBRecipeHandler.py
+++ b/cellpack/autopack/DBRecipeHandler.py
@@ -217,10 +217,9 @@ class CompositionDoc(DataDoc):
 
         # check for cycles in the graph
         if len(upload_order) != len(in_degree):
-            raise ValueError("Cyclic dependency detected in compositions.")
+            raise ValueError("Circular dependency detected in compositions.")
 
         # reverse the order since we need inner nodes first
-        # TODO: refactor to return the correct order
         upload_order.reverse()
         return upload_order
 


rugeli:
TODO: troubleshoot `upload_composition` to detect any changes in inner compositions and create new top level compositions.

i.e. updating the `count` of `A` should create a new comp `A` and a new comp `space`. These new compositions should be linked to the corresponding recipe as references.
```
 "composition": {
        "space": {
            "regions": {
                "interior": [
                    "A"
                ]
            }
        },
        "A": {
            "object": "sphere_25",
            "count": 1
        }
    }
```

rugeli:
> TODO: troubleshoot `upload_composition` to detect any changes in inner compositions and create new top level compositions.
> 
> i.e. updating the `count` of `A` should create a new comp `A` and a new comp `space`. These new compositions should be linked to the corresponding recipe as references.
> 
> ```
>  "composition": {
>         "space": {
>             "regions": {
>                 "interior": [
>                     "A"
>                 ]
>             }
>         },
>         "A": {
>             "object": "sphere_25",
>             "count": 1
>         }
>     }
> ```

Solution:
We've implemented topological sorting to make sure that the compositions are processed in the correct dependency order, from innermost to outermost, with proper firebase reference replacement. So in the case above, a field in `A` is modified, the system will detect the change and create a new `A`, a new `space` and a new `recipe`.

meganrm (approved):

ascibisz (approved):
LGTM!

PR #326, opened 2025-04-04
Fix/upload nested object

rugeli:
Problem
=======
A quick fix for remote recipes.
Alli and I were testing more recipes with the uploading system and found two bugs:
- `priority` is not being uploaded to composition documents
- nested object in composition is not being replaced by its reference  

Solution
========
- added `priority` to default values in compost ion
- fixed the `replace_region_references` function to resolve nested region references    
- update unittests 
with @Alli

## Type of change

Steps to Verify:
----------------
1. Upload a recipe to Firebase and check the data

### DIFF
--- a/cellpack/tests/test_comp_doc.py
+++ b/cellpack/tests/test_comp_doc.py
@@ -11,6 +11,7 @@ def test_composition_doc_as_dict():
         object={"test_key": "test_value"},
         regions={},
         molarity=None,
+        priority=None,
     )
     expected_dict = {
         "name": "test",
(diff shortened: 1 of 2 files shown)


rugeli:
Hey @Alli, let me know if you spot anything else that needs fixing in this PR. After you've had a look, I'll add a second reviewer to also take a look before merging. Thanks!

cellpack/tests/test_comp_doc.py
```diff
@@ -11,13 +11,15 @@ def test_composition_doc_as_dict():
         object={"test_key": "test_value"},
         regions={},
         molarity=None,
+        priority=None,
```
ascibisz, line 14: nit: should we put a value in for priority for one of these tests? instead of just None?

mogres (approved):

PR #332, opened 2025-05-01
Fix: default db for packing remote configs

rugeli:
Problem
=======
we need to set the default db to staging when running batch jobs

Solution
========
- similar to how we did to recipe, added default db to config file packings
with @alli

## Type of change

Steps to Verify:
----------------
1. upload a config to firebase: upload -c [CONFIG-LOCAL-PATH] to `staging`
2. pack a firebase recipe with a firebase config files: pack -r [RECIPE-FIREBASE-PATH] -c [CONFIG-FIREBASE-PATH] -d
3. verify if it uses `staging` db

### DIFF
--- a/cellpack/autopack/__init__.py
+++ b/cellpack/autopack/__init__.py
@@ -398,7 +398,7 @@ def load_file(
     if is_remote_path(filename):
         database_name, file_path = convert_db_shortname_to_url(filename)
         db = DATABASE_IDS.handlers().get(database_name)
-        initialize_db = db()
+        initialize_db = db(default_db="staging") if use_docker else db()
 
         if not initialize_db._initialized:
             readme_url = "https://github.com/mesoscope/cellpack?tab=readme-ov-file#introduction-to-remote-databases"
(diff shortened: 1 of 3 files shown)


rugeli:
@ascibisz Alli you may see the credential error again `raise S3UploadFailedError(
boto3.exceptions.S3UploadFailedError: Failed to upload out/one_sphere/jitter/results_one_sphere_default-run_1.0.0_seed_0.simularium to cellpack-results/simularium/results_one_sphere_default-run_1.0.0_seed_0.simularium: An error occurred (ExpiredToken) when calling the PutObject operation: The provided token has expired.
WARNING: All log messages before absl::InitializeLog() is called are written to STDERR
E0000 00:00:1746136053.744972 25009560 init.cc:232] grpc_wait_for_shutdown_with_timeout() timed out.`

I did the unsetting steps and run `aws sts assume-role  --role-arn arn:aws:iam::771753870375:role/batch  --role-session-name Test`. It went through.

PR #336, opened 2025-05-14
Feature/docker file for ecs

ascibisz:
Problem
=======
As part of our desire to speed up packings on the cellpack AWS server, we started testing working with ECS, instead of Batch. This requires a different Docker image to be built for us to run on ECS. 

Solution
========
Created a new Dockerfile and docker-entrypoint for the image we'll have on ECS, as well as a basic python HTTP server for the client to send packings to.

For now, I want to keep both our ECS and our Batch Dockerfiles, because I think Batch can still have some really good use cases in the future. I moved everything into a /docker subdirectory and changed some file names, just to keep things organized and clear.

Also updated the README with instructions for building and running both docker images.


docker/server.py
```diff
@@ -0,0 +1,32 @@
...
+
+async def pack_handler(request: web.Request) -> web.Response:
+    recipe = request.rel_url.query.get("recipe")
+    if recipe is None:
+        # pack request must include "recipe" query parameter
+        raise web.HTTPBadRequest()
+    config = request.rel_url.query.get("config")
+    job_id = str(uuid.uuid4())
+    os.environ["AWS_BATCH_JOB_ID"] = job_id
+    pack(recipe=recipe, config_path=config, docker=True)
```
rugeli, line 19: should we add an exception here to prevent server crash if the pack function fails?

rugeli (approved):

Issue #355, opened 2025-07-29
comprehensive recipe validation during loading


rugeli:
## Use Case
*Please provide a use case to help us understand your request in context*
As we support more customized recipes through cellpack client, we need a comprehensive validation system for recipe fields. This will help catch mistyped/missing values early and give users with cleaner error messages.  

## Solution
*Please describe your ideal solution*
Introduce a validation layer during loading phase, it should:
- validate the presence and type of fields 
- provide helpful error messages 
- (TBD)validate nested structure

PR #368, opened 2025-08-25
Add initial Pydantic schema for recipe validation

rugeli:
Problem
=======
kickstarting of #355 

Solution
========
- added starting schema for recipe validation
- formatted pydantic validation error message to be more readable 
- added CLI command to validate a single recipe  
- added optional validation step before packing

The current validation covers:
- recipe metadata 
- objects
- gradients 
- composition 

The schema is based on our documentation, existing test pattern and my understanding of relationships between fields. We can always expand coverage by adding more fields and new tests.  

Next validations we can work on but need further discussion:
- version, format version (how many decimal places? what version are we on? what's the rule for version update?)
- config
...

## Question:
1. in the gradient data below, "nucleus" is a reference to a mesh, but following the validation rule that all the object references should exist in the `objects` section. Do we want to update the reference to `mean_nucleus` instead? or allow `mode_settings.object` to also reference to a composition key? 
```
  "gradients": {
        "surface_gradient": {
            "description": "gradient based on distance from the surface of a mesh",
            "weight_mode": "exponential",
            "pick_mode": "rnd",
            "mode": "surface",
            "mode_settings": {
                "object": "nucleus", // ---> causes error 
                "scale_to_next_surface": false
            },
            "weight_mode_settings": {
                "decay_length": 0.05
            }
        }
    },
```
in terminal:
```
validate examples/recipes/v2/pcna_surface_gradient.json
2025-08-25 15:33:15 | root | ERROR | validate:27 | validate() | Validation errors found:
 Value error, gradients.surface_gradient.mode_settings.object references 'nucleus' which does not exist in objects section
```

## Type of change

Steps to Verify:
----------------
1. to validate a recipe: `validate [recipe-path]`
2. to validate before packing: `pack -r [recipe-path] -v`
3. let me know if there are errors that are missed to catch

Note: if the validate command can't be found, try running `pdm install`

Screenshots (optional):
-----------------------
error example
Screenshot 2025-08-25 at 2 32 42 PM


mogres:
> in the gradient data below, "nucleus" is a reference to a mesh, but following the validation rule that all the object references should exist in the objects section. Do we want to update the reference to mean_nucleus instead? or allow mode_settings.object to also reference to a composition key?

I would suggest we allow `mode_settings.object` to refer to a composition key. Items in the composition can refer to the same base object but have different properties (e.g., a sphere base object can show up as an endosome and peroxisome both in the composition section with different counts, colors, etc.).

rugeli:
@mogres main updates since your last comment:
- added a new rule to the validation schema: gradient references must point to either existing objects or compositions(string key) 
- set validate default to `True` in `pack.py`, so recipe data is validated before packing
Screenshot 2025-09-05 at 9 46 38 AM

let me know if there's anything else we'd like to include in this pull request

cellpack/autopack/validation/recipe_models.py
```diff
@@ -0,0 +1,467 @@
...
+
+    jitter_attempts: int = Field(5, ge=1)
+    max_jitter: ThreeFloatArray = Field([1, 1, 1], min_length=3, max_length=3)
+    rotation_range: Optional[float] = Field(None, ge=0)
+    rotation_axis: Optional[Union[ThreeFloatArray, None]] = None
+    use_rotation_axis: Optional[bool] = None
+    principal_vector: Optional[ThreeFloatArray] = Field(
+        None, min_length=3, max_length=3
+    )
+    orient_bias_range: Optional[List[float]] = Field(None, min_length=2, max_length=2)
```
mogres, line 184: Can add limits from [-pi, pi] in a future PR

mogres (approved):
Looks good and the validation seems to work as expected!! Thanks Ruge!

Recording some next steps that came to my mind for future PRs:
1. Adding expected ranges for all recipe fields
2. Adding tests for recipe validation
3. Removing duplicated validation steps from other places in the code

PR #381, opened 2025-09-08
upload outputs to s3

rugeli:
Problem
=======
addresses the first step of implementing download results feature for the web client 
issue link: https://github.com/mesoscope/cellpack-client/issues/70

Solution
========
- added a general upload directory function in `AWSHandler`
- added an outputs-specific upload function in `DBRecipeHandler`
- ensure each job run creates its own folder for uploading outputs (approx. 11-20 files per run)
- update `job_status` doc in firebase with the outputs s3 folder path

## Type of change

Steps to Verify:
----------------
1. `pack -r cellpack/tests/recipes/v2/test_spheres.json -c cellpack/tests/packing-configs/test_config.json -d`(when we test locally the job_id is `None`, the system will get it when we are actually running jobs on the server)

Screenshots (optional):
-----------------------
Screenshot 2025-09-08 at 12 18 36 PM

### DIFF
--- a/cellpack/bin/pack.py
+++ b/cellpack/bin/pack.py
@@ -82,8 +82,9 @@ def pack(
         job_id = os.environ.get("AWS_BATCH_JOB_ID", None)
         if job_id:
             handler = DATABASE_IDS.handlers().get(DATABASE_IDS.AWS)
+            # temporarily using demo bucket before permissions are granted
             initialized_handler = handler(
-                bucket_name="cellpack-results",
+                bucket_name="cellpack-demo",
                 sub_folder_name="runs",
                 region_name="us-west-2",
             )


cellpack/autopack/DBRecipeHandler.py
```diff
@@ -561,6 +563,76 @@ def upload_job_status(self, job_id, status, result_path=None, error_message=None
...
+            error_msg = f"Failed to upload packing results to S3: {e}"
+            logging.error(error_msg)
+            return {"success": False, "error": error_msg}
+
+    def _update_outputs_directory(job_id, outputs_directory):
+        # Update firebase with the new outputs directory
+        handler = DATABASE_IDS.handlers().get("firebase")
+        initialized_db = handler(default_db="staging")
+        if job_id:
+            initialized_db.update_or_create(
```
ascibisz, line 625: nit: should we update the timestamp field in the job_Status collection here too?

cellpack/bin/pack.py
```diff
@@ -69,6 +86,36 @@ def pack(recipe, config_path=None, analysis_config_path=None, docker=False):
...
+    :param recipe_name: Name of the recipe being packed
+    """
+    try:
+        if job_id:
+            output_path = Path(output_folder)
+            if not output_path.exists():
+                log.error(f"Output folder does not exist: {output_folder}")
+                return
+
+            uploader = DBUploader(db_handler=None)
```
ascibisz, line 111: I understand that we're initializing `DBUploader` with `db_handler=None` because we need to use 2 different db_handlers in the process of upload_outputs_to_s3 (one for S3 and one for firebase), but it feels like a weird pattern to establish. Like perhaps it would be better to do the s3 uploading with an s3 db_handler and if that succeeds, initialize a new DBUploader with a firebase db_handler to update the logs?

We're on a time crunch so this probably isn't critical, but could be worth adding a TODO comment or something to come back and clean it up?

PR #386, opened 2025-09-10
Updates to make python 3.11 work for our ECS Docker image

ascibisz:
It's time to deploy a new docker image to ECS with python 3.11, since that's what cellpack uses now.

Ran into a couple bugs along the way, some of which were not related to the upgrade, and fixing them all here. 

The Docker image generated by this updated Dockerfile has already been uploaded to ECR and is currently being used in out ECS instance!

### DIFF
--- a/cellpack/autopack/__init__.py
+++ b/cellpack/autopack/__init__.py
@@ -177,8 +177,8 @@ REPLACE_PATH = {
     "autoPACKserver": autoPACKserver,
     "autopackdir": autopackdir,
     "autopackdata": appdata,
-    DATABASE_IDS.GITHUB + ":": autoPACKserver,
-    DATABASE_IDS.FIREBASE + ":": None,
+    f"{DATABASE_IDS.GITHUB.value}:": autoPACKserver,
+    f"{DATABASE_IDS.FIREBASE.value}:": None,
 }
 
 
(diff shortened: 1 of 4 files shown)


cellpack/autopack/__init__.py
```diff
@@ -177,8 +177,8 @@ def checkPath():
     "autoPACKserver": autoPACKserver,
     "autopackdir": autopackdir,
     "autopackdata": appdata,
-    f"{DATABASE_IDS.GITHUB}:": autoPACKserver,
-    f"{DATABASE_IDS.FIREBASE}:": None,
+    DATABASE_IDS.GITHUB + ":": autoPACKserver,
+    DATABASE_IDS.FIREBASE + ":": None,
```
ascibisz, line 181: I'm imagining this is somehow related to the python version change, but the string formatting wasn't working correctly and these keys were just getting set to "DATABASE_IDS.GITHUB:" and "DATABASE_IDS.FIREBASE:", which was causing errors. This change makes it so the keys are "github:" and "firebase:", as expected

docker/server.py
```diff
@@ -64,7 +64,7 @@ async def init_app() -> web.Application:
     app.add_routes(
         [
             web.get("/hello", server.hello_world),
-            web.post("/pack", server.pack_handler),
+            web.post("/start-packing", server.pack_handler),
```
ascibisz, line 67: this is the path the API gateway sends requests to, `/pack` is out of date

PR #388, opened 2025-09-11
fix: add upload_results to test configs

rugeli:
# Problem

test users without database credentials are blocked by the simularium result upload error

# Solution

option 1 (in this pr): include `upload_results: false` in the config file
- pros: we get to keep the result tab pop-out feature since it relies on a publicly accessible url
- cons: test configs must always include the option explicitly

option 2: default disable results upload
- pros: users without remote setups can run tests without being blocked 
- cons: results need to be opened manually from the local file in simularium viewer

option 3: use the cellpack web app to view recipe data, see embedded simularium result, and download output files. Without any pre setups!

Steps to Verify:
----------------
1. v1: `pack -r examples/recipes/v1/NM_Analysis_FigureB1.0.json -c examples/packing-configs/run.json`
2. v2:  `pack -r examples/recipes/v2/one_sphere.json -c examples/packing-configs/run.json`
3. Pack from remote server: `pack -r  github:recipes/NM_Analysis_FigureB1.0.json  -c examples/packing-configs/run.json`

## Type of change

### DIFF
--- a/cellpack/autopack/__init__.py
+++ b/cellpack/autopack/__init__.py
@@ -388,6 +388,12 @@ def load_file(
 ):
     if is_remote_path(filename):
         database_name, file_path = convert_db_shortname_to_url(filename)
+        if database_name == DATABASE_IDS.GITHUB:
+            # right now we support github files as json
+            local_file_path = get_local_file_location(
+                file_path, destination=destination, cache=cache, force=force
+            )
+            return json.load(open(local_file_path, "r")), "github", False
         db = DATABASE_IDS.handlers().get(database_name)
         initialize_db = db(default_db="staging") if use_docker else db()
 


rugeli:
## Packing flow recap:
### we can pack from
    - local recipe (json)
    - github recipe (json)
    - firebase recipe (firebase doc, requires firebase read access)
    
### when `upload_results = true` in config (default):
    - pack a json (local/github): pack --> ask for s3 and firebase creds --> upload result to s3 and save url to firebase --> auto opens viewer from the public url 
    - pack a firebase recipe: ask for which db (requires firebase creds) --> pack --> ask for s3 access --> upload result to s3 and save url to firebase --> auto opens viewer from the public url 

### when `upload_results = false` in config:
    - pack a json (local+github): pack --> results saved to a local folder --> manually open/drag simularium file into the viewer **(works for users with access to the python package)**
    - pack a firebase recipe: ask for which db (requires firebase creds) --> pack --> results saved locally --> manual open/drag 

### core conflicts
- generates a sharable url for simularium result
- automatically opens the viewer
vs
- accessible to all users without requiring cloud setups 
       
### solution
- pack via web app: select a recipe --> pack on server --> view results in embedded viewer 
- adjust `upload_results` config option accordingly 
- (experimental) blob the local file to a blob:url and pass it to the viewer, currently failing because simularium viewer rejects blob:url in `trajUrl`

mogres (approved):
Really helpful update to unblock users without s3/firebase credentials!! I like your solution to direct users to the web app!

Some thoughts to follow up on:
1. We should update the default config options in `config_loader.py` to set `upload_results` to `False` for users running the recipe without a config. Ideally, we should also update the readme to run recipes without a config file at all and just have it use the defaults from `config_loader` to reduce any setup friction. Maybe these changes could become a part of this PR.
2. If a user attempts to pack a recipe with `upload_results: true` or tries to pack a firebase recipe without credentials setup, we catch the exception and log a message to the output with a link to the website. Currently the function `upload_results` doesn't catch `S3UploadFailedError` that is raised, resulting in unintended logs.
3. We should prune the list of example recipes and configs to only keep the ones that we think are helpful. This would involve updating/removing the peroxisome example recipes. 

Lmk what you think!

ascibisz (approved):

PR #390, opened 2025-09-26
Feature/validate firebase recipes

rugeli:
# Problem

# Solution

## Type of change


ascibisz (approved):
I tested it with all our cellpack client example recipes and it works great!

PR #391, opened 2025-09-26
Update deepdiff to version 8.6.1

mogres:
# Problem

Fixes [security vulnerability](https://github.com/mesoscope/cellpack/security/dependabot/66) in deepdiff. 
# Solution

Updated deepdiff to version `8.6.1`

## Type of change

## Change summary:

Add new deepdiff version to `pyproject.toml` and update `uv.lock`

## Steps to Verify:

Please verify if deepdiff related operations still work after this update

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,7 +20,7 @@ requires-python = "~=3.11.0"
 dependencies = [
     "aicsimageio",
     "boto3",
-    "deepdiff",
+    "deepdiff>=8.6.1",
     "fire",
     "firebase-admin",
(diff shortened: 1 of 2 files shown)


mogres:
> Good catch on dependency versions! 
All with some help from dependabot 😄 

> Since we’ve switched to hash-based deduplication for recipe uploads, it’s safe to remove `deepdiff` now. I had missed cleaning it up earlier. Thanks for pointing it out!

If it's no longer a required dependecy, is it okay if I go ahead and remove it in this PR?

PR #402, opened 2025-10-02
validate recipes during load

### LINKED ISSUE #401: Validate v1/v2/remote recipe data in `recipe_loader`
## Use Case

with the merge of #390, we can now properly validate v2 and firebase recipes. But packing a v1 example recipe (e.g. `pack -r examples/recipes/v1/NM_Analysis_FigureB1.0.json`) throws an error, because validation currently runs before v1 recipe data is converted to the unified format.  

## Proposed Solution

- move validation call from `pack` to `recipe_loader`, after all recipe data is normalized
- make sure CLI `validate` command goes through the same path, so we can run `validate [V1/V2/firebase-RECIPE-PATH]` as expected

rugeli:
# Problem

closes #401 

# Solution

- move validation to `RecipeLoader._read()` after data migration
    - fixed validation error when packing v1 recipes
    - consolidated validation to one spot that works for all recipes (v1/v2/firebase)
- Simplify `validate.py` to use RecipeLoader for all recipe
- improved data migration tests for `orientBiasRotRange`

## Type of change

* Refactor (non-breaking change which adds functionality)

## Steps to Verify:
1. run validate [ANY-VALID-RECIPE-PATH]
    - v1 `validate examples/recipes/v1/one_sphere.json`
    - v2 `validate examples/recipes/v2/pcna_surface_gradient.json`
    - firebase `validate firebase:recipes/gradients_v_default -d `
2. run pack -r [ANY-VALID-RECIPE-PATH]

### DIFF
--- a/cellpack/autopack/loaders/migrate_v1_to_v2.py
+++ b/cellpack/autopack/loaders/migrate_v1_to_v2.py
@@ -63,16 +63,34 @@ def get_representations(old_ingredient):
 
 
 def convert_rotation_range(old_ingredient):
-    range_min = (
-        old_ingredient["orientBiasRotRangeMin"]
-        if "orientBiasRotRangeMin" in old_ingredient
-        else -pi
-    )
-    range_max = (
-        old_ingredient["orientBiasRotRangeMax"]
-        if "orientBiasRotRangeMax" in old_ingredient
-        else pi
-    )
+    has_min = "orientBiasRotRangeMin" in old_ingredient
+    has_max = "orientBiasRotRangeMax" in old_ingredient
+
+    if has_min and has_max:
+        range_min = old_ingredient["orientBiasRotRangeMin"]
+        range_max = old_ingredient["orientBiasRotRangeMax"]
+        if range_min > range_max:
+            range_min, range_max = range_max, range_min
+    elif has_min and not has_max:
+        provided_min = old_ingredient["orientBiasRotRangeMin"]
+        if provided_min > pi:
+            range_min = -pi
+            range_max = provided_min
+        else:
+            range_min = provided_min
+            range_max = pi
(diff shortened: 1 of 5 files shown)


ascibisz (approved):
I really like the code change moving the validation to recipe_loader! I tested it out and it works great.

I don't fully understand the context of the changes in migrate_v1_to_v2, but the code looks good and is very readable. But just wanted to call out that blindspot I have, I don't know much about migrate_v1_to_v2 and if you have concerns about that code change, I would suggest tagging another reviewer who has more context.

cellpack/autopack/loaders/migrate_v1_to_v2.py
```diff
@@ -63,16 +63,34 @@ def get_representations(old_ingredient):
...
+        if provided_max < -pi:
+            range_min = provided_max
+            range_max = pi
+        else:
+            range_min = -pi
+            range_max = provided_max
+    else:
+        range_min = -pi
+        range_max = pi
+
```
rugeli, line 93: fixed a bug in the conversion function to prevent invalid ranges where `min>max` (e.g. `[6, pi]`)

cellpack/tests/test_recipe_version_migration.py
```diff
@@ -332,7 +332,7 @@ def test_convert_v1_to_v2(
                         "type": INGREDIENT_TYPE.SINGLE_SPHERE,
                     },
                     "B": {
-                        "orient_bias_range": [6, pi],
+                        "orient_bias_range": [-pi, 6],
```
rugeli, line 335: updated test data for `orient_bias_range` accordingly

mogres (approved):
Love the recipe validation during read time! I have some comments about the orientation bias range conversion, but the validation updates look good to me!

PR #403, opened 2025-10-02
clean up and update V2 recipes and configs in example and tests folder

### LINKED ISSUE #399: clean up example recipes and configs
## Use Case

some of the recipes and configs in the current example folder are outdated or less maintained

## Proposed Solution

- go through the existing examples and trim down to just the ones we think are useful
- update or remove the peroxisome example recipes in particular
- make sure the remaining ones pass validation and ready to pack

rugeli:
# Problem

closes #399 

# Solution

- removed not in used v2 recipes and configs
- fixed a ploty error we came cross when `show_sphere_trees` set to `true` in config
- added recipes from the manuscript as examples
- v1 recipes and gitignored `converted` folders are remaining the same for now since they are less used 

with @mogres


mogres:
TODO:

- [ ] Add screenshots of example recipes in the README file

cellpack/autopack/Analysis.py
```diff
@@ -1179,7 +1179,7 @@ def doloop(
             across seeds
         """
         number_of_packings = packing_config_data["number_of_packings"]
-        plot_figures = packing_config_data.get("save_plot_figures", True)
+        plot_figures = packing_config_data["save_plot_figures"]
```
meganrm, line 1182: what's the reason for this change?

meganrm (approved):

cellpack/autopack/Analysis.py
mogres: We added a default entry for `save_plot_figures` in `config_loader.py` (it did not exist before). The pattern now matches how other settings are updated from the config.

PR #417, opened 2025-10-22
Feature/client upload script

ascibisz:
# Problem
Add upload script for uploading recipes that we want to show up on the cellpack client site
[Link to story or ticket](https://github.com/AllenCell/cellpack-client/issues/113)

# Solution
I decided it makes more sense to have this upload script in this repo rather than the cellpack-client repo so we could reuse the `upload` python code rather than re-implementing it in typescript.

Also added some documentation to explain how to use this code to upload the necessary metadata for the cellpack client.

### DIFF
--- a/cellpack/bin/upload.py
+++ b/cellpack/bin/upload.py
@@ -58,9 +58,9 @@ def upload(
 
     :return: void
     """
-    if studio and (not name or not recipe_path or not config_path or not fields):
+    if studio and (not name or not recipe_path or not fields):
         sys.exit(
-            "To upload a recipe for cellPACK studio, please provide --name, --recipe_path, --config_path, and --fields arguments."
+            "To upload a recipe for cellPACK studio, please provide --name, --recipe_path, and --fields arguments."
         )
 
     recipe_id = ""


examples/client-data/example_editable_fields.json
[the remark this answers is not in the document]
ascibisz: We could definitely add `step_size` and change the terminology to `decay_length`! But those changes also need to happen on the cellpack-client site side of things, currently the client wouldn't read either of those two fields, so I would be hesitant to update this example documentation until those changes are actually in effect on the client side.

cellpack/bin/upload_to_client.py
```diff
@@ -0,0 +1,60 @@
...
+            config_data = ConfigLoader(config).config
+            config_id = db_handler.upload_config(config_data, config)
+        if fields:
+            editable_fields_data = json.load(open(fields, "r"))
+            for field in editable_fields_data.get("editable_fields", []):
+                id, _ = db_handler.upload_data("editable_fields", field)
+                editable_fields_ids.append(id)
+        recipe_metadata = {
+            "name": name,
+            "recipe": recipe_id,
```
meganrm: can these be changed to either `display_name` and `id` (are these nested in an object called 'recipe_metadata'?) or `recipe_display_name` and `recipe_id` without causing too many problems? or should this be a separate PR?

rugeli (approved):
I tested the new changes and they worked well! Just have a few ideas for non-critical refinements. I agree we can move forward with this pr and follow up with improvements after the submission.
- when params like `config`, `db_id`, `name` aren't set, the metadata ends up with null values, which may affect frontend display. add warning/error handing, making these options required or add some frontend checks?
- Ruge's todo: FirebaseHandler shouldn't be imported here, this is a missed spot from the previous handler refactor. we can do an overall cleanup later

docs/STUDIO_SITE.md
```diff
@@ -0,0 +1,14 @@
...
+* Production site: https://cellpack.allencell.org/
+* GitHub repository: https://github.com/AllenCell/cellpack-client
+
+### Publishing Recipes to Studio Site
+To add new example recipes to the client site, the recipe and its associated metadata must be uploaded to Firebase.
+
+1. Set up access to the staging Firebase account, following [these instructions](https://github.com/mesoscope/cellpack/blob/main/docs/REMOTE_DATABASES.md#firebase-firestore)
+2. Determine which recipe fields should be editable in the client, and create a JSON file specifying these editable fields following the schema outlined in [this example](https://github.com/mesoscope/cellpack/blob/0f140859824086d73edab008ff381b5e5717db8b/examples/client-data/example_editable_fields.json)
+3. Follow cellPACK installation requirements [in README](https://github.com/mesoscope/cellpack?tab=readme-ov-file#installation)
+4. Run the following script to upload necessary files to Firebase: `upload -s -r <path_to_recipe_file> -c <path_to_config_file> -f <path_to_editable_fields_file> -n "Name of Recipe" -o <path_to_simularium_result_file>`
```
rugeli: nit: It might help to update the parameter description to something like <local_path_to_simularium_result_file> to make it clearer that it expects a path such as `out/one_sphere/spheresSST/results_one_sphere_debug_1.0.0_seed_0.simularium`

meganrm (approved):

PR #419, opened 2025-10-22
Server should Accept Recipe JSON 

ascibisz:
# Problem
Currently, when a user packs an edited recipe via the cellpack client, we upload the edited recipe to firebase, pass that reference to the server, and then the server retrieves the recipe from firebase. To minimize firebase calls and improve efficiency of the client, we're changing that flow to instead send the edited recipe JSON as the body of the packing request to the server. 

We also wanted a way to check if we have already packed a recipe and if we have, return that result file rather than running the whole packing again. To do this, we are calculating a hash for the recipe objects before they are packed. The hash for each recipe will be uploaded to firebase along with its result files, so we can query firebase for a given hash to see if that recipe has a packing result already.

### 1. **Deduplication & Caching**
- **BEFORE**: Each request generated a unique UUID, no deduplication possible
- **AFTER**: JSON recipes generate deterministic hash, enabling job deduplication

### 2. **Input Flexibility & Backwards Compatibility**
- **BEFORE**: Only recipe file paths supported via query parameter
- **AFTER**: Supports both recipe file paths AND direct JSON recipe objects in request body, plus optional config parameter

### 3. **Smart Job Management**
- **BEFORE**: Generated UUID for each job without deduplication, every request creates new job regardless of content
- **AFTER**: Uses deterministic hash for JSON recipes, enabling job reuse for identical recipes

### 4. **Firebase Request Reduction**
- **BEFORE**: Every edited recipe was uploaded to firebase by the client and downloaded from firebase by the server
- **AFTER**: Edited recipes are passed in the body of the packing request, so no firebase uploads or downloads occur

### 5: **Unified Results Upload**
- **BEFORE**: Simularium result file was uploaded to S3 twice per job, once on its own and once as part of the full output files upload
- **AFTER**: Only upload Simularium result file once by keeping track of its path when we upload all output files

### New Server Components:
1. **`DataDoc.generate_hash()`** - Creates deterministic hash from recipe JSON
2. **`job_exists()`** - Checks if job already completed in Firebase
3. **Enhanced request handling** - Reads JSON from request body
4. **Smart job ID generation** - Uses hash for JSON recipes, UUID for file paths

### Request Flow Changes:
1. **Input validation** now checks both query params and request body
2. **Hash-based deduplication** for JSON recipes
3. **Backward compatibility** maintained for file-based recipes
4. **Consistent job tracking** with hash parameter

## Benefits

1. **Reduced Server Load**: Identical recipes don't reprocess
2. **Faster Client Response**: Instant return for duplicate JSON requests  
3. **Better Resource Utilization**: No redundant compute for same recipes
4. **Improved API Design**: JSON recipes easier for programmatic access
5. **Reduced Firebase Usage**: Passing recipe directly instead of uploading to firebase

# CellPACK Server Job Workflow Changes

## BEFORE: Original Server Workflow
```mermaid
graph TD
    A[Client Request] --> B[POST /start-packing]
    B --> C{Check for recipe URL param}
    C -->|Missing| D[Return 400 Error]
    C -->|Present| E[Generate UUID for job_id]
    E --> F[Create Background Task]
    F --> G[Return job_id immediately]
    F --> I[Initiate packing]
    I --> J[Load recipe from firebase<br>using file path from<br>URL param]
    J --> K[Execute packing]
    K --> L{Packing succeeds?}
    L -->|Success| M[S3: Upload outputs to S3<br>Firebase: Update job status to SUCCEEDED]
    L -->|Failure| N[Firebase: Update job status to FAILED]
    
    style A fill:#e1f5fe
    style G fill:#c8e6c9
    style M fill:#fff3e0
    style N fill:#ffcdd2
```

## AFTER: Enhanced Server Workflow with JSON Recipe Support
```mermaid
graph TD
    A[Client Request] --> B[POST /start-packing]
    B --> C{Check inputs}
    C -->|No recipe - no URL param<br>and no request body| D[Return 400 Error]
    C -->|Has recipe path URL param| E[Generate UUID for job_id]
    C -->|Has recipe JSON in request body| F[Generate hash from JSON]
    F --> G{Packing result exists<br>in firebase for this hash?}
    G -->|Yes| H[Return existing hash<br>as job_id]
    G -->|No| I[Use hash as job_id]
    E --> J[Create Background Task]
    I --> J
    J --> K[Return job_id immediately]
    J --> L[Initiate packing]
    L --> M{Input type?}
    M -->|Recipe path| N[Load recipe from firebase<br>using file path from<br>URL param]
    M -->|JSON body| O[Load recipe from JSON dict<br>from request body]
    N --> P[Execute packing]
    O --> P
    P --> Q{Packing succeeds?}
    Q -->|Success| R[S3: Upload outputs to S3<br>Firebase: Update job status to SUCCEEDED]
    Q -->|Failure| S[Firebase: Update job status to FAILED]
    
    style A fill:#e1f5fe
    style K fill:#c8e6c9
    style R fill:#fff3e0
    style S fill:#ffcdd2
    style G fill:#ffeb3b
    style H fill:#c8e6c9
```

### DIFF
--- a/docker/Dockerfile.ecs
+++ b/docker/Dockerfile.ecs
@@ -4,7 +4,7 @@ WORKDIR /cellpack
 COPY . /cellpack
 
 RUN python -m pip install --upgrade pip --root-user-action=ignore
-RUN python -m pip install . --root-user-action=ignore
+RUN python -m pip install . -r requirements/linux/requirements.txt --root-user-action=ignore
 
 EXPOSE 80
 


docker/server.py
```diff
@@ -12,41 +10,61 @@ class CellpackServer:
...
+        else:
+            body = None
+        if not recipe and not body:
             raise web.HTTPBadRequest(
                 "Pack requests must include recipe as a query param"
             )
         config = request.rel_url.query.get("config")
-        job_id = str(uuid.uuid4())
+
+        dedup_hash = DataDoc.generate_hash(body)
```
ascibisz: check if this causes an error in the case that recipe was passed in as a firebase path

docker/server.py
```diff
@@ -12,41 +10,61 @@ class CellpackServer:
...
-        os.environ["AWS_BATCH_JOB_ID"] = job_id
-        self.update_job_status(job_id, "RUNNING")
+    def _get_firebase_handler(self, database_name="firebase"):
+        handler = DATABASE_IDS.handlers().get(database_name)
+        initialized_db = handler(default_db="staging")
+        if initialized_db._initialized:
+            return initialized_db
+        return None
+
+    def job_exists(self, dedup_hash):
```
ascibisz: might be worth adding a comment describing this function

cellpack/autopack/DBRecipeHandler.py
ascibisz: Major changes here:

1. We realized the results collection in firebase is not necessary. For cellpack studio purposes, it is redundant with job_status and we only look to job_status. For locally run packings, we can skip uploading the result path to firebase at all and just directly open the simularium file. This allowed us to remove the whole `ResultDoc` class, the `DBMaintenance` class,  and remove `upload_result_metadata`
2. We're now using `dedup_hash` instead of `job_id` as the id for server packings
3. Refactored `upload_job_status` to handle the functionality of `update_outputs_directory` since they were basically the same, and removed `update_outputs_directory`

cellpack/bin/cleanup_tasks.py
ascibisz: We realized the results collection in firebase is not necessary. For cellpack studio purposes, it is redundant with job_status and we only look to job_status. For locally run packings, we can skip uploading the result path to firebase at all and just directly open the simularium file. This allowed us to remove the cleanup code for the `results` collection (which was the only firebase cleanup we were doing in this repo anyways)

.github/workflows/cleanup-firebase.yml
ascibisz: We realized the results collection in firebase is not necessary. For cellpack studio purposes, it is redundant with job_status and we only look to job_status. For locally run packings, we can skip uploading the result path to firebase at all and just directly open the simularium file. This allowed us to remove the cleanup code for the results collection (which was the only firebase cleanup we were doing in this repo anyways)

docker/server.py
```diff
@@ -12,41 +11,66 @@ class CellpackServer:
...
                 "Pack requests must include recipe as a query param"
             )
         config = request.rel_url.query.get("config")
-        job_id = str(uuid.uuid4())
+
+        if body:
+            dedup_hash = DataDoc.generate_hash(body)
+            if self.job_exists(dedup_hash):
+                return web.json_response({"jobId": dedup_hash})
+            job_id = dedup_hash
```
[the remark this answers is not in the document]
ascibisz: we validate this during the packign process

cellpack/bin/pack.py
```diff
@@ -25,23 +24,32 @@
...
+    :param hash: string argument, dedup hash identifier for tracking/caching results
 
     :return: void
     """
     packing_config_data = ConfigLoader(config_path, docker).config
 
-    recipe_loader = RecipeLoader(
-        recipe, packing_config_data["save_converted_recipe"], docker
-    )
+    if isinstance(recipe, dict):
```
ascibisz: Try moving this to recipe_loader, it might be cleaner

cellpack/autopack/loaders/recipe_loader.py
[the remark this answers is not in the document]
mogres: One suggestion to handle multiple input streams:

1. Update the `RecipeLoader` init to accept a single `input_data` argument which can be a path to a file or a dictionary. This should avoid potetntial bugs from conflicting inputs passed through `input_file_path` and `json_recipe` since there is only one input now.
2. Handle resolving the path or dict inside the `read()` method. This might require getting rid of the file path and file extension attributes which are not really used I think

cellpack/autopack/upy/simularium/simularium_helper.py
```diff
@@ -1385,64 +1385,33 @@ def raycast(self, **kw):
...
         simularium_file = Path(f"{file_name}.simularium")
-        url = None
-        job_id = os.environ.get("AWS_BATCH_JOB_ID", None)
-        file_name, url = simulariumHelper.store_result_file(
-            simularium_file, storage="aws", batch_job_id=job_id
-        )
-        if file_name and url:
-            simulariumHelper.store_metadata(
-                file_name, url, db="firebase", job_id=job_id
+        if dedup_hash is None:
```
mogres, line 1390: Can you clarify what happens here if `dedup_hash` is not None? Is the result not opened in browser if this is run locally?

cellpack/autopack/writers/__init__.py
```diff
@@ -196,8 +196,11 @@ def save_as_simularium(self, env, seed_to_results_map):
         number_of_packings = env.config_data.get("number_of_packings", 1)
         open_results_in_browser = env.config_data.get("open_results_in_browser", False)
         upload_results = env.config_data.get("upload_results", False)
+        dedup_hash = getattr(env, "dedup_hash", None)
```
mogres: Default return for `getattr` is `None`
```suggestion
        dedup_hash = getattr(env, "dedup_hash")
```

cellpack/autopack/DBRecipeHandler.py
```diff
@@ -529,42 +497,34 @@ def upload_config(self, config_data, source_path):
...
-                    "status": str(status),
-                    "result_path": result_path,
-                    "error_message": error_message,
-                },
-            )
+            db_handler = self.db
+            # If db is AWSHandler, switch to firebase handler for job status updates
+            if hasattr(self.db, "s3_client"):
+                handler = DATABASE_IDS.handlers().get(DATABASE_IDS.FIREBASE)
+                db_handler = handler(default_db="staging")
```
mogres, line 519: Would this overwrite the `self.db` attribute? Might be better to create a new handler object or use `.copy()` just in case

cellpack/autopack/DBRecipeHandler.py
```diff
@@ -529,42 +497,34 @@ def upload_config(self, config_data, source_path):
...
-                {
-                    "timestamp": timestamp,
-                    "status": str(status),
-                    "result_path": result_path,
-                    "error_message": error_message,
-                },
-            )
+            db_handler = self.db
+            # If db is AWSHandler, switch to firebase handler for job status updates
+            if hasattr(self.db, "s3_client"):
```
mogres: Is it possible to directly check if this is an `AWSHandler` using `isinstance`? Idk if it is possible for it to not have the `s3_client` attribute but just in case.

cellpack/autopack/DBRecipeHandler.py
```diff
@@ -661,8 +624,11 @@ def upload_outputs_to_s3(self, output_folder, recipe_name, job_id):
                     f"{base_url}/{file_info['s3_key']}"
                     for file_info in upload_result["uploaded_files"]
                 ]
+                simularium_url = None
+                for url in public_urls:
+                    if url.endswith(".simularium"):
+                        simularium_url = url
```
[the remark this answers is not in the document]
mogres: Nice!

cellpack/autopack/DBRecipeHandler.py
```diff
@@ -890,23 +838,4 @@ def compile_db_recipe_data(db_recipe_data, obj_dict, grad_dict, comp_dict):
...
-        Check if the results in the database are expired and delete them if the linked object expired.
-        """
-        self.result_doc.handle_expired_results()
-
-    def readme_url(self):
-        """
-        Return the URL to the README file for the database setup section.
-        """
-        return "https://github.com/mesoscope/cellpack?tab=readme-ov-file#introduction-to-remote-databases"
+DB_SETUP_README_URL = "https://github.com/mesoscope/cellpack?tab=readme-ov-file#introduction-to-remote-databases"
```
mogres: Can this move to the top of the file?

cellpack/bin/pack.py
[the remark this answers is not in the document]
mogres: Might be good to have an integration test to check if the changes in this file don't break the local workflow: `pack -r RECIPE_PATH -c CONFIG_PATH` with a mock recipe and config. Not sure if we have this already

cellpack/bin/pack.py
```diff
@@ -52,6 +60,7 @@ def pack(
     autopack.helper = helper
     env = Environment(config=packing_config_data, recipe=recipe_data)
     env.helper = helper
+    env.dedup_hash = hash
```
mogres, line 58: You might have to update the environment `__init__` to initialize the `dedup_hash` attribute as `None`

cellpack/bin/pack.py
```diff
@@ -78,24 +87,22 @@ def pack(
...
-            )
-            uploader = DBUploader(db_handler=initialized_handler)
-            uploader.upload_packing_results_workflow(
-                source_folder=env.out_folder,
-                recipe_name=recipe_data["name"],
-                job_id=job_id,
-                config_data=packing_config_data,
-                recipe_data=recipe_loader.serializable_recipe_data,
-            )
+    if docker and hash:
```
mogres, line 86: Can you add a short docstring saying that this branch runs if `pack` is called from cellPACK studio (i.e. `docker` and `hash` are both provided)?

docker/Dockerfile.ecs
```diff
@@ -4,7 +4,7 @@ WORKDIR /cellpack
 COPY . /cellpack
 
 RUN python -m pip install --upgrade pip --root-user-action=ignore
-RUN python -m pip install . -r requirements/linux/requirements.txt --root-user-action=ignore
+RUN python -m pip install . --root-user-action=ignore
```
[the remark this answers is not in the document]
mogres: Might be worth looking into. AFAIK `aiohttp` is installed as a subdependency and listed explicitly in the `requirements.txt` file. @rugeli, this also relates to your comment about exporting `requirements.txt` files where we may want to have a list of pinned dependencies to install directly using `pip` without having to resolve from `pyproject.toml` every time.

docker/server.py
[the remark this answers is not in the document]
mogres: > We check this dedup hash against those existing in the job_status firebase collection to see if this exact recipe has already been packed, and if we find a match, we return it rather than running the packing

Just to clarify, do we still run the packing even if the dedup hash exists in firebase? Or do we just pull the results?

mogres (commented):
Read through the changes and left some clarifying comments. I might do some testing later just to verify none of the other workflows are breaking. 

Really awesome work putting this together! Excited to see the speedup in cellPACK studio.

cellpack/autopack/upy/simularium/simularium_helper.py
rugeli: For server jobs, the result file is uploaded as part of the outputs directory via `upload_outputs_to_s3`, the frontend can then fetch the result_path and display simularium result in Studio, so we intentionally skip the separate upload when dedup_hash is not None here.
For local packings, dedup_hash is None, so the existing upload +open in browser is preserved.

cellpack/autopack/writers/__init__.py
rugeli: good catch! fixed.

cellpack/autopack/DBRecipeHandler.py
rugeli: to your question above: no, `self.db` is not overwritten. In this case, `db_handler` is reassigned to a local variable that points to the Firebase handler, `self.db` remains as `s3_client`. 

And I agree, `isinstance` would be more explicit and robust here, thanks for calling that out!

rugeli:
> We check this dedup hash against those existing in the job_status firebase collection to see if this exact recipe has already been packed, and if we find a match, we return it rather than running the packing

> Just to clarify, do we still run the packing even if the dedup hash exists in firebase? Or do we just pull the results?

@mogres Yeah in that case, we skip the packing entirely and return the `debup_hash` as jobId immediately without creating a new packing. This is only applies to server-initiated packings with a JSON body, local packings go through the CLI are unaffected.

docker/Dockerfile.ecs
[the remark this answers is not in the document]
rugeli: @mogres Makes sense! for the matter of scope, I reverted the change back to use requirements.txt. I'll add this to my TODO. Thanks!

rugeli:
> Thank you for the update and the great effort @rugeli and @ascibisz!! PR looks good to me.
> 
> Future follow-up:
> 
> 1. Can we add tests to check if the same dedup hash is returned for recipes that have the same entries but in a different order? I believe lists for example are not updated when the json is dumped with sorted keys. I can make a PR for this if you'd like.

Thanks! @mogres I can definitely add the tests. I'm already working on some broader integration test improvements for #460, so I can include this there and request your review once the pr is up

Issue #422, opened 2025-10-27
Use github CLI for PR comments and remove `iterative/setup-cml` action


rugeli:
## Use Case
the CI currently runs `iterative/setup-cml` action for posting PR comments and generating reports. Since this action is less maintained and starts causing issues, we need to remove it and use the native github CLI instead.

## Proposed Solution

replace all instances of `iterative/setup-cml` with github CLI commands for PR commenting and reporting

PR #434, opened 2025-11-21
Fix `Comment` job failing on `main` pushes 

rugeli:
# Problem

#422 merged but broke the `Comment` job on `main`
The `iterative/setup-cml` action quietly skipped commenting when running on a push to main, but the raw gh commands assume a pull request always exists. 

# Solution

- Added check guards so the `Comment` job only runs when there is a pull request

## Type of change


rugeli:
@ascibisz Alli I’m going to go ahead and merge this with one approval since the action is breaking on main. let me know if you have any thoughts or feedback afterward! thanks!

PR #435, opened 2025-11-25
Studio: add recipe and config json to outputs for downloading 

rugeli:
# Problem

closes https://github.com/AllenCell/cellpack-client/issues/103

# Solution

- added recipe and config data to output folder so users can download them from UI. they will appear as `recipe.json` and `config.json`

Note: we send in the full `packing_config_data` to output folder as a json, regardless of whether the pack used default configs or a config file

## Type of change


rugeli:
> I tried running this locally and it works as expected for local config and recipe files, but it doesn't work for firebase config and recipe files (it just tries to read the firebase path as a local file path). Maybe we can discuss potential solutions at pair programming?

darn I keep forgetting about the firebase path handling, thanks for catching it! I'll get that fixed in this pr before tomorrow.

PR #458, opened 2026-03-18
admin/requirements-update_2026-03-18_17-14_23257457346

mogres:
Updating requirements.txt.

### DIFF
--- a/requirements/linux/requirements.txt
+++ b/requirements/linux/requirements.txt
@@ -5,60 +5,56 @@ alabaster==1.0.0
     # via sphinx
 annotated-types==0.7.0
     # via pydantic
-anyio==4.10.0
+anyio==4.12.1
     # via httpx
 appnope==0.1.4 ; sys_platform == 'darwin'
     # via ipykernel
-asttokens==3.0.0
+asttokens==3.0.1
     # via stack-data
-babel==2.17.0
+babel==2.18.0
     # via sphinx
-bioio==3.0.0
+bioio==3.2.0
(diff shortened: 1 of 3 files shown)


rugeli (commented):
Come to think of it, what's the use case for these files now that we use `uv` and `pyproject.toml`? Are we keeping `requirements.txt` around so the codebase remains pip-compatible for users who don't have `uv` installed? If that's helpful for users, then it makes sense to keep them, otherwise we could consider deprecating them gradually.

mogres:
> Come to think of it, what's the use case for these files now that we use `uv` and `pyproject.toml`? Are we keeping `requirements.txt` around so the codebase remains pip-compatible for users who don't have `uv` installed? If that's helpful for users, then it makes sense to keep them, otherwise we could consider deprecating them gradually.

Good point @rugeli! The `requirements.txt` files are not strictly necessary as `pip install` works with `pyproject.toml`. However, I think some of our docker workflow and CI uses the `pip install` workflow reading from the requirements files. How about this:

0. Merge this PR for now to see if that addresses any of the reported dependabot dependency issues. 
1. We look for and update instances of `pip install` from requirementes files in the codebase and in the CI to use `uv sync` or `pip install` from `pyproject.toml` instead
2. We stop tracking the `requirements.txt` files, and get rid of the automated action that exports them after dependencies change. 
3. Update `README.md` to reflect the new installation instructions

rugeli (approved):
Sounds like a plan!

PR #462, opened 2026-04-14
Test updates: stabler debup hash and cli integration coverage

### LINKED ISSUE #460: have an integration test to check if the changes in this file don't break the local workflow
Might be good to have an integration test to check if the changes in this file don't break the local workflow: `pack -r RECIPE_PATH -c CONFIG_PATH` with a mock recipe and config. Not sure if we have this already

_Originally posted by @mogres in https://github.com/mesoscope/cellpack/pull/419#discussion_r2968271923_

rugeli:
# Problem

follow-up to #419, closes #460 

# Solution

- added `_normalize_for_hashing` to sort lists whose element order is not meaningful (e.g.  `"interior": ["A", "B", "C"]`, `"interior": [ {"object": "green_sphere", "count": 5}, "outer_sphere"]` ), while leaving order-sensitive lists(vectors, colors, etc.) unchanged.
- added integration tests in `test_pack_cli.py`
- refactored aws tests to write temporary files to `tmp_path` instead of the working dir

One note: this change can alter hash values for recipes that have lists, so re-run the same recipe after this merge may hash differently. Shouldn't cause issues, but worth noting so we can keep an eye on database tidiness.

## Type of change

* Refactor (non-breaking change which fixes an issue)

### DIFF
--- a/cellpack/tests/test_aws_handler.py
+++ b/cellpack/tests/test_aws_handler.py
@@ -31,7 +31,7 @@ def test_get_aws_object_key():
         assert object_key == "test_folder/test_file"
 
 
-def test_upload_file():
+def test_upload_file(tmp_path):
     with mock_aws():
         aws_handler = AWSHandler(
             bucket_name="test_bucket",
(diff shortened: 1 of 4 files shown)


cellpack/tests/test_aws_handler.py
```diff
@@ -63,17 +63,17 @@ def test_create_presigned_url():
             Bucket="test_bucket",
             CreateBucketConfiguration={"LocationConstraint": "us-west-2"},
         )
-        with open("test_file.txt", "w") as file:
-            file.write("test file")
-        aws_handler.upload_file("test_file.txt")
+        test_file = tmp_path / "test_file.txt"
+        test_file.write_text("test file")
+        aws_handler.upload_file(str(test_file))
```
rugeli, line 68: a bit out of scope but a simple change, switched tests to write via `tmp_path` so they leave no trace like `test_file.txt` in the repo root

cellpack/tests/test_pack_cli.py
```diff
@@ -0,0 +1,82 @@
...
+    """
+    `pack()` also accepts a recipe dict directly, so
+    the docker server can forward a parsed JSON body without writing it to
+    db first.
+    """
+    config_path = _write_config(tmp_path)
+    pack(recipe=recipe_data, config_path=str(config_path))
+
+
+def test_pack_with_default_config(tmp_path, monkeypatch):
```
rugeli: similar to `tmp_path` in aws tests, we don't want test files written inside the repo, so we change directory to `tmp_path` with `monkeypatch.chdir()`

meganrm (approved):
looks good, thanks for all the tests

mogres (approved):
Very clever solution using `._is_positional`!! Looks good to me and thanks for adding tests (and fixing the test file issue)!

PR #464, opened 2026-04-15
Remove Support for Recipe URL Param

### LINKED ISSUE #463: Cleanup Cellpack Server
Now that we've fully deployed cellpack-client to only pass recipe as JSON body instead of as a firebase path in the URL, we should update `server.py` to only accept recipe as a JSON body.

We had previously accepted both to stay backwards compatible during the deployment process, but now we're done deploying!

Follow up to [this PR](https://github.com/mesoscope/cellpack/pull/419)

ascibisz:
# Problem
When we started supporting a JSON body for a packing request to the server, we wanted to stay backwards compatible and also accept the recipe path as a URL parameter.

Now that we have fully deployed cellpack studio with the change to pass the recipe in the JSON body, we can strip out the support for the URL param way of sending a recipe from server.py

We also don't actually use AWS batch for cellpack server, so I also stripped out the references to batch and the dockerfile we had used in batch


docker/server.py
rugeli: I'm wondering how linter didn't catch this unused import. We can remove?

rugeli (approved):
only one unused `uuid` import, otherwise looks great to me!

PR #467, opened 2026-04-27
Maint/deprecate requirements txt

### LINKED ISSUE #459: Deprecate requirements.txt files
> Come to think of it, what's the use case for these files now that we use `uv` and `pyproject.toml`? Are we keeping `requirements.txt` around so the codebase remains pip-compatible for users who don't have `uv` installed? If that's helpful for users, then it makes sense to keep them, otherwise we could consider deprecating them gradually.

Good point @rugeli! The `requirements.txt` files are not strictly necessary as `pip install` works with `pyproject.toml`. However, I think some of our docker workflow and CI uses the `pip install` workflow reading from the requirements files. How about this:

0. Merge this PR for now to see if that addresses any of the reported dependabot dependency issues. 
1. We look for and update instances of `pip install` from requirementes files in the codebase and in the CI to use `uv sync` or `pip install` from `pyproject.toml` instead
2. We stop tracking the `requirements.txt` files, and get rid of the automated action that exports them after dependencies change. 
3. Update `README.md` to reflect the new installation instructions

_Originally posted by @mogres in https://github.com/mesoscope/cellpack/issues/458#issuecomment-4094434733_

rugeli:
# Problem

closes #459 

# Solution

- removed the per-platform `requirements.txt` files and the workflow that generates them
- updated `Dockerfile.ecs` to use `uv sync` against `uv.lock`, so dependency versions stay pinned  
- added a new `[dependency-groups].server = ["aiohttp"] ` group to declare the web-server dependency
- updated documentation

## Type of change

* Maintenance (non-breaking change which preserves existing functionality)

## Steps to Verify:
                                                                                                                                                                               
  1. pull the branch and run `uv sync  --group server`, it should resolve and install `aiohttp` along with the rest of the deps from `uv.lock` 
  2. (optional) run`docker build -f docker/Dockerfile.ecs -t cellpack-ecs`, the image should build                                                                  
  3. (optional) run`docker run -p 80:80 cellpack-ecs` then go to `http://0.0.0.0:80/hello`, should return `Hello from the cellPACK server`


docker/Dockerfile.ecs
```diff
@@ -4,7 +4,9 @@ WORKDIR /cellpack
 COPY . /cellpack
 
 RUN python -m pip install --upgrade pip --root-user-action=ignore
-RUN python -m pip install . -r requirements/linux/requirements.txt --root-user-action=ignore
+RUN python -m pip install uv --root-user-action=ignore
+RUN uv sync --no-default-groups --extra server --frozen
+ENV PATH="/cellpack/.venv/bin:$PATH"
```
rugeli, line 9: initially I used `pip install ".[server]"`, but that resolves fresh from PyPI on each build which means rebuilds could pick up different versions. Switching to `uv sync` keeps the install pinned to `uv.lock`, giving us more consistent builds.

docs/CONTRIBUTING.md
```diff
@@ -15,12 +15,11 @@ Ready to contribute? Here's how to set up `cellpack` for local development.
...
     ```
 
-3. Install the project in editable mode by following installation instructions [here](./docs/INSTALL.md). (It is also recommended to work in a virtual environment):
+3. Install the project in editable mode by following installation instructions [here](./INSTALLATION.md). (It is also recommended to work in a virtual environment):
 
     ```bash
     cd cellpack/
-    pip install -r requirements/linux/requirements.txt
-    pip install -e .[dev]
+    uv sync
```
rugeli, line 22: to address copilot's review comment(slightly out of scope): `pip install -e .[dev]` has been silently failing on `main` since the `dev` lives in `[dependency-groups]` and pip's `[extra]` syntax can't read. Switched the contributor setup to use `uv sync` instead.

docker/Dockerfile.ecs
ascibisz: just tried running this locally and it worked perfectly!

ascibisz (approved):

mogres (approved):
Looks great -- thanks!!

PR #470, opened 2026-06-01
fix KeyError: 'weight_mode' when packing v2.1 recipes with under-specified gradients

### LINKED ISSUE #465: Validation schema improvements
## Use Case

@mogres ran into friction while packing recipes that expected to be valid. Pydantic validation rejected the recipe during pack/validate, which blocking the workflow.

It would be helpful if @mogres could drop a short note here with:
- which recipe you were packing 
- the validation error message you saw
- what shape/format/value was expected to be accepted 
- (optional)any feedback, requests, suggestions for improvement

That context will help us decide whether this is a schema that's too strict, a normalization gap in `recipe_loader`, or something else in the workflow.
 
## Proposed Solution

TBD based on user feedback

rugeli:
# Problem

closes #465 

Sorry for the delay on this one, after tracing it back through the packing pipeline, I figured the gradient packing error is not a problem with our Pydantic schema. The root cause is a data-pipeline gap. Gradient defaults were only ever filled in one place: the v2.0→v2.1 migration. But `recipe_loader._read` only runs migration when format_version != current_version. A recipe hand-authored directly as 2.1 skips migration entirely, so its gradients reach `Gradient.__init__` still missing keys.

In short, without the changes in this pr, relabeling `format_version` as `2.0`, everything else stays the same, the recipe `cellpack/tests/recipes/v2/test_peroxisome_combined_gradient.json` would pass.

# Solution

in `recipe_loader`:
- added gradient normalization that converts list-of-dict form and fill missing top-level keys. Made this conversion happen before validation.
- removed now-redundant gradient dict to list block 

with @mogres 

Note: If we want to further refactor, we can:
1. make use of the validator output. `validate_recipe` builds a normalized copy that `_read` currently discards, we currently only use validator for validating which makes sense to me. But if we want to, we can make the validated recipe source of truth. This will be a larger pr so I'd keep this separate.
2. unify version migration scripts(default filling, structure flatten&nested, etc). more files involved, so better be another separate pr  

## Type of change

## Steps to Verify:
1. run `pack -r cellpack/tests/recipes/v2/test_peroxisome_combined_gradient.json -c cellpack/tests/packing-configs/test_gradient_mixing.json`


meganrm (approved):

mogres (approved):
Thanks Ruge! I ran the verification successfully.

